PR #1, opened 2020-02-18
[WIP] Add contents to dask-geopandas

jsignell:
This PR sets up the minimal version of dask-geopandas.

To Do:
 - [x] Create classes for `GeometryArray`, `GeoSeries` and `GeoDataFrame`
 - [ ] Add special geopandas methods to these classes
 - [ ] Add special geopandas properties to these classes - (is there a generic way of implementing this?)
 - [ ] Write dask implementations for methods that don't just map to each partition.
 - [ ] Test
 
Not To Do Yet:
 - [ ] Implement spatial methods following patterns in spatialpandas
 - [ ] Add dask-geopandas IO functions to lazily load from geo files


dask_geopandas/dataframe.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @property
+    @derived_from(geopandas.base.GeoPandasBase)
+    def bounds(self):
+        return self.map_partitions(
+            getattr,
+            "bounds",
+            token=self._name + "-bounds",
+            meta=self._meta.bounds
+        )
```
jsignell: Is there a better way of putting properties onto the new classes?

jsignell: @TomAugspurger @mrocklin

jsignell: Basically I was just starting to write a `_bind_property` classmethod and wondering if such a thing already existed. I'm still working out how to set a property without knowing the name ahead of time, but I'm sure it's possible.

TomAugspurger: Not identical, but `dask.dataframe.Accessor._delegate_property_get` may provide inspiration: https://github.com/dask/dask/blob/632f68c04154915da84c0e0befba1391a583f596/dask/dataframe/accessor.py#L16-L44

jsignell: Hmm. That is helpful, but since the property in this case is on the class itself, if we try to catch `getattr` on the fly, we get circular references. Ok I will try to see if I can wed this approach with the bind approach.

jsignell: Ok I think I've written a reasonable generic version of a property binder. Now I just need to choose whether to pass a list of properties to add or whether to bind all the properties that exist. The list approach is safer.

jsignell:
The next thing for me to figure out is how `make_meta` and `meta_nonempty` work. I can also start adding methods.

TomAugspurger:
@jsignell https://github.com/geopandas/geopandas/pull/1143/files may be helpful there. I *think* that `meta_nonempty` eventually uses `make_array_nonempty`, but I may be wrong.

jsignell:
@TomAugspurger I think I should be able to register a custom `meta_nonempty` like this:

https://github.com/rapidsai/cudf/blob/8d7bf34ff045fbe1760c90ef825e6719bab1ff80/python/dask_cudf/dask_cudf/backends.py#L13-L16

So I might not need a change in geopandas as far as I can tell.

jsignell:
Oh I think I understand. I can just steal the bits that you are registering in that PR.

dask_geopandas/array.py
```diff
@@ -0,0 +1,50 @@
...
+    y = da.asarray(y)
+    if z is not None:
+        z = da.asarray(z)
+    out = _points_from_xy(x, y, z)
+    aout = da.empty(len(x))
+    aout[:] = out
+    return GeometryArray.from_array(aout)
+
+
+class GeometryArray(da.core.Array, OperatorMethodMixin):
```
jorisvandenbossche: Just wondering: what is the reason that you also need to create a dask version of `GeometryArray` ?

My assumption was that only GeoSeries/GeoDataFrame would be needed, since those would be consisting of (geo)pandas chunks that use geopandas' GeometryArray extension array.

dask_geopandas/dataframe.py
```diff
@@ -0,0 +1,112 @@
...
+            lambda x: getattr(x, "total_bounds"),
+            token=self._name + "-total_bounds",
+            meta=self._meta.total_bounds,
+            aggregate=agg,
+        )
+
+    @property
+    def unary_union(self):
+        # Would need a dask representation of each shapely.geometry
+        raise NotImplementedError
```
jorisvandenbossche: I am probably not familiar enough with dask's internals, but you can't do this as a reduction in the sense of the unary_union of the unary_union of each partition? (like you would do the sum of a dask series)

dask_geopandas/array.py
jsignell: Yeah, I think it's not actually needed. I just kind of stubbed it out in the first pass (I should mention that this branch is very WIP). I originally thought that points_from_xy would be a good first entrypoint to support.

dask_geopandas/dataframe.py
jsignell: Oh! That seems reasonable yes. I was trying to capture which properties could use `_bind_properties` and do something with the others.

I'd rather have all the attrs that the geopandas class has. And any that haven't been thought about could just raise NotImplemented. That way it is clear what work remains to be done.

dask_geopandas/array.py
jorisvandenbossche: Yeah, I think `points_from_xy` *is* a good entrypoint, but probably can also return a GeoSeries directly to avoid needing a partitioned array.

jsignell: Ok good idea :)

dask_geopandas/core.py
```diff
@@ -88,7 +92,19 @@ def crs(self):
     @crs.setter
     def crs(self, value):
         """Sets the value of the crs"""
-        raise NotImplementedError
+        self.dask = self.set_crs(value).dask
```
mrocklin: I think that you will also need to update some other state like `._name`

dask_geopandas/core.py
```diff
@@ -88,7 +92,19 @@ def crs(self):
...
+        self._meta.crs = value
+
+        def set_crs(df, crs):
+            df = df.copy(deep=False)
+            df.crs = crs
+            return df
+
+        return self.map_partitions(
+            set_crs, value, meta=self._meta, enforce_metadata=False
+        )
```
mrocklin: What happens if you don't supply `_meta` here?  Ideally we would then call `set_crs` to `self._meta` which would be good because then `self.crs` (which presumably calls `self._meta.crs`) would have the right coordinate metadata.

dask_geopandas/core.py
```diff
@@ -88,7 +92,19 @@ def crs(self):
...
-        raise NotImplementedError
+        self.dask = self.set_crs(value).dask
+
+    def set_crs(self, value):
+        self._meta.crs = value
+
+        def set_crs(df, crs):
+            df = df.copy(deep=False)
+            df.crs = crs
+            return df
```
mrocklin: This can be changed later, but we typically avoid using dynamically created functions like this, and instead try to use top level functions.  This is useful because dynamically generated functions are a bit more expensive and complex to serialize.

dask_geopandas/core.py
```diff
@@ -0,0 +1,248 @@
...
+
+        return self.map_partitions(
+            set_crs, value, meta=self._meta, enforce_metadata=False
+        )
```
jsignell: From @mrocklin:

> Ideally _meta would also be affected by map_partitions. Hopefully you wouldn't need to specify it ahead of time. That way ddf.crs would be the actual coordinate data
Also my guess is that you'll have to update other things than .dask
Maybe this conversation should move to an issue or PR somewhere for longevity?

jsignell: Yes. I was having trouble because I hadn't properly registered nonempty metas for geoseries and geodataframe. Now that I have, things are working as you suggest.

mrocklin:
Woot

dask_geopandas/core.py
jsignell: Ah ok I'd seen that pattern and just hadn't thought about the rationale. I'll move it out.

dask_geopandas/core.py
jsignell: Yep turns out I needed `_meta`

dask_geopandas/core.py
```diff
@@ -0,0 +1,248 @@
...
+        such as an authority string (eg "EPSG:4326") or a WKT string.
+        """
+        return self._meta.crs
+
+    @crs.setter
+    def crs(self, value):
+        """Sets the value of the crs"""
+        new = self.set_crs(value)
+        self._meta = new._meta
+        self.dask = new.dask
```
mrocklin: I strongly suspect that you also need to change `_name`

jsignell: Oh yes I see now. That makes sense. Do you think this kind of property setting is a good idea? Or should I just raise a `NotImplemented` that tells people to use `set_crs`

mrocklin: I think that if we can do it safely then yes, it's good to support this kind of behavior.  It's just rare enough that it's error prone as we build the functionality.  That's what review is for  though :)

dask_geopandas/core.py
```diff
@@ -0,0 +1,330 @@
...
+
+    _partition_type = geopandas.base.GeoPandasBase
+
+    def __repr__(self):
+        s = "<dask_geopandas.%s | %d tasks | %d npartitions>"
+        return s % (type(self).__name__, len(self.dask), self.npartitions)
+
+    def to_dask_dataframe(self):
+        """Create a dask.dataframe object from a dask_geopandas object"""
+        return self.map_partitions(M.to_pandas)
```
jorisvandenbossche: `map_partitions` calls the method on each partition, i.e. the underlying geopandas.GeoDataFrame/Series right? 
Because geopandas has no `to_pandas` method right now.

dask_geopandas/core.py
```diff
@@ -0,0 +1,330 @@
...
+    "convex_hull",
+    "envelope",
+    "exterior",
+    "interiors",
+    "bounds",
+]:
+    _Frame._bind_property(name)
+
+for name in [
+    "geometry",
```
jorisvandenbossche: We might want to bind "geometry" to the GeoDataFrame as well, to have similar behaviour for geopandas, where "geometry" always returns the "active geometry column", whathever the actual name of the column (although this can certainly be confusing behaviour as well ..)

dask_geopandas/core.py
```diff
@@ -0,0 +1,330 @@
...
+    "overlaps",
+    "touches",
+    "within",
+]:
+    meth = getattr(geopandas.base.GeoPandasBase, name)
+    GeoSeries._bind_elemwise_comparison_method(
+        name, meth, original=geopandas.base.GeoPandasBase
+    )
+    GeoDataFrame._bind_elemwise_comparison_method(
+        name, meth, original=geopandas.base.GeoPandasBase
```
jorisvandenbossche: Why do those need to be bound to both GeoSeries and GeoDataFrame (instead of once to `_Frame`)

dask_geopandas/core.py
```diff
@@ -0,0 +1,330 @@
...
+    def set_crs(self, value):
+        """Set the value of the crs on a new object"""
+
+        def set_crs(df, crs):
+            df = df.copy(deep=False)
+            df.crs = crs
+            return df
+
+        return self.map_partitions(set_crs, value, enforce_metadata=False)
+
```
jorisvandenbossche: We probably want to add a `to_crs` as well here? (that should be a relatively easy "map_partitions" call I think, but also needs to update meta)

(can also be done in a follow-up PR)

dask_geopandas/core.py
```diff
@@ -0,0 +1,330 @@
...
+                    concatted[3::4].max(),  # maxy
+                )
+            )
+
+        return self.reduction(
+            lambda x: getattr(x, "total_bounds"),
+            token=self._name + "-total_bounds",
+            meta=self._meta.total_bounds,
+            aggregate=agg,
+        )
```
jorisvandenbossche: In principle we know the shape of the returned array here. That might be something to specify to dask? (now you get a shape of `(nan, )`

dask_geopandas/core.py
```diff
@@ -0,0 +1,330 @@
...
+    @property
+    @derived_from(geopandas.base.GeoPandasBase)
+    def total_bounds(self):
+        def agg(concatted):
+            return np.array(
+                (
+                    concatted[0::4].min(),  # minx
+                    concatted[1::4].min(),  # miny
+                    concatted[2::4].max(),  # maxx
+                    concatted[3::4].max(),  # maxy
```
jorisvandenbossche: We just switched to this in geopandas as well -> better to use `np.nanmax(..)` instead of `.max()`, because if some partitions only have empty/missing geometries, we want to skip those and get the total bounds of the non-empty geometries

dask_geopandas/core.py
```diff
@@ -0,0 +1,330 @@
...
+    @property
+    def sindex(self):
+        """Need to figure out how to concatenate spatial indexes"""
+        raise NotImplementedError
+
+    @property
+    @derived_from(geopandas.base.GeoPandasBase)
+    def unary_union(self):
+        attr = "unary_union"
+        meta = GeometryCollection()
```
jorisvandenbossche: I think BaseGeometry is better here, since eg unary union of points gives MultiPoint, which is not a GeometryCollection subclass

(not sure if that actually matters though)

dask_geopandas/core.py
```diff
@@ -0,0 +1,330 @@
...
+        new = self.set_crs(value)
+        self._meta = new._meta
+        self.dask = new.dask
+
+    def set_crs(self, value):
+        """Set the value of the crs on a new object"""
+
+        def set_crs(df, crs):
+            df = df.copy(deep=False)
+            df.crs = crs
```
jorisvandenbossche: We probably want to check that the `crs` is None (not already set), and in that case raise a warning or error (maybe with a keyword to force it). 

Otherwise I think people will accidentally call this assuming their geometries get transformed.

(maybe for a follow-up though; we should also align with a similar method in GeoPandas)

jsignell:
@jorisvandenbossche thanks for looking!

> Thoughts on merging now / soon ?

I'd like to merge it soon. Let me take another pass on Monday and pick off any of the easy stuff and then I'll merge and convert any outstanding comments to issues.

jorisvandenbossche:
Sounds good!

dask_geopandas/backends.py
```diff
@@ -0,0 +1,54 @@
...
+import geopandas
+from geopandas.array import GeometryDtype, from_shapely
+
+from .core import GeoSeries, GeoDataFrame
+
+get_parallel_type.register(geopandas.GeoDataFrame, lambda _: GeoDataFrame)
+get_parallel_type.register(geopandas.GeoSeries, lambda _: GeoSeries)
+
+
+@make_meta.register((GeometryCollection, BaseGeometry))
```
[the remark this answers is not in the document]
jsignell: Seems like registering BaseGeometry is indeed enough.

dask_geopandas/backends.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+@make_scalar.register(GeometryDtype.type)
+def _(x):
+    return shapely.geometry.Point(0, 0)
+
+
+@meta_nonempty.register(geopandas.GeoSeries)
+def _nonempty_geoseries(x, idx=None):
+    s = _nonempty_series(x, idx)
```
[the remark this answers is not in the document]
jsignell: Oh good point!

dask_geopandas/core.py
jsignell: Yeah I copied that from dask_cudf, but I don't understand how it could ever work. I am just going to delete it for now.

dask_geopandas/core.py
jsignell: Ok that makes sense. I have added it now following the pattern of `crs`.

dask_geopandas/core.py
jsignell: Binding to `_Frame` works fine. I think this is just a copy-paste hold-out from dask-cudf.

dask_geopandas/core.py
jsignell: This one I think we should do as a follow-up for sure.

jsignell: #3

dask_geopandas/core.py
jsignell: sounds good

dask_geopandas/core.py
jsignell: #4

dask_geopandas/core.py
jsignell: ok!

jsignell:
I'll merge this if it passes.

dask_geopandas/core.py
```diff
@@ -1,11 +1,28 @@
...
+    df = df.copy(deep=False)
+    df.crs = crs
+    return df
+
+
+def _finalize(results):
+    if isinstance(results[0], (geopandas.GeoSeries, geopandas.GeoDataFrame)):
+        output = pd.concat(results)
+        output.crs = results[0].crs
+        return output
```
jorisvandenbossche: Ah, this might actually be a geopandas bug then that concatting GeoSeries does not preserve the CRS (will look into that)

dask_geopandas/core.py
jorisvandenbossche: This seems to have sneaked in again in a later commit?

dask_geopandas/core.py
jorisvandenbossche: It indeed is -> https://github.com/geopandas/geopandas/issues/1340

jorisvandenbossche:
Woohoo! Thanks again for getting this started!

dask_geopandas/core.py
jsignell: Oh cool. I included the fix in this PR anyways, but we can take it out later.

PR #9, opened 2020-06-05
Add basic example notebook

jorisvandenbossche:
It doesn't show / explain much, yet, but to have at least some example code.


mrocklin:
Nice.

PR #13, opened 2020-08-12
ENH: add Parquet IO

jorisvandenbossche:
Addressing the Parquet IO part of https://github.com/jsignell/dask-geopandas/issues/6

For now this is heavily reusing dask's Parquet machinery, which is easy thanks to the Engine classes (I don't know if this will keep being possible long term, though, if we want to add more customizations, eg related to spatial partitioning etc ..)

(still need to add tests, but this is working nicely for basic example!)

### DIFF
--- a/dask_geopandas/io/parquet.py
+++ b/dask_geopandas/io/parquet.py
@@ -1,0 +1,49 @@
+from functools import partial
+from typing import TYPE_CHECKING
+
+import pandas as pd
+
+import geopandas
+from geopandas.io.arrow import _geopandas_to_arrow, _arrow_to_geopandas
+
+import dask.dataframe as dd
+from dask.dataframe.io.parquet.arrow import ArrowEngine
+
+if TYPE_CHECKING:
+    import pyarrow
+
+
+class GeoArrowEngine(ArrowEngine):
+
+    @classmethod
+    def read_metadata(cls, *args, **kwargs):
+        meta, stats, parts, index = super().read_metadata(*args, **kwargs)
+
+        # Update meta to be a GeoDataFrame
+        # TODO convert columns based on GEO metadata (this will now only work for a default "geometry" column)
+        meta = geopandas.GeoDataFrame(meta)
+
+        return (meta, stats, parts, index)
+
+    @classmethod
+    def _arrow_table_to_pandas(
+        cls, arrow_table: "pyarrow.Table", categories, **kwargs
+    ) -> pd.DataFrame:
+        _kwargs = kwargs.get("arrow_to_pandas", {})
+        _kwargs.update({"use_threads": False, "ignore_metadata": False})
+
+        # TODO support additional keywords
+        return _arrow_to_geopandas(arrow_table)
+
+    @classmethod
+    def _pandas_to_arrow_table(
+        cls, df: pd.DataFrame, preserve_index=False, schema=None
+    ) -> "pyarrow.Table":
+        # TODO add support for schema
+        table = _geopandas_to_arrow(df, index=preserve_index)
+        # table = pa.Table.from_pandas(df, preserve_index=preserve_index, schema=schema)
+        return table
+
+
+to_parquet = partial(dd.to_parquet, engine=GeoArrowEngine)
+read_parquet = partial(dd.read_parquet, engine=GeoArrowEngine)
--- a/dask_geopandas/__init__.py
+++ b/dask_geopandas/__init__.py
@@ -8,6 +8,8 @@ from .core import (
     from_geopandas,
     from_dask_dataframe,
 )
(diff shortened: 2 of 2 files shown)


jorisvandenbossche:
Aso, this needs a custom branch of dask right now -> https://github.com/dask/dask/pull/6505

jorisvandenbossche:
> Aso, this needs a custom branch of dask right now -> dask/dask#6505

I added some extra compat code, so that for now, this doesn't need the dask branch, but will just work with released dask.

Also added a basic test.

.pre-commit-config.yaml
```diff
@@ -3,10 +3,10 @@ repos:
     rev: stable
     hooks:
       - id: black
-        language_version: python3.7
+        language_version: python3
```
jorisvandenbossche, line 6: I otherwise ran into troubles running pre-commit (actually installing the envs) from my py3.8 dev environment

dask_geopandas/io/parquet.py
```diff
@@ -0,0 +1,126 @@
...
+        _kwargs.update({"use_threads": False, "ignore_metadata": False})
+
+        # TODO support additional keywords
+        return _arrow_to_geopandas(arrow_table)
+
+    @classmethod
+    def _pandas_to_arrow_table(
+        cls, df: pd.DataFrame, preserve_index=False, schema=None
+    ) -> "pyarrow.Table":
+        from geopandas.io.arrow import _geopandas_to_arrow
```
jsignell, line 48: Why is the import in the method? Can we move it out and wrap it in a try except depending on pyarrow availablitity

dask_geopandas/io/parquet.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # TODO add support for schema
```
jsignell, line 50: Should we raise an error if schema is provided in the meantime?

dask_geopandas/io/parquet.py
```diff
@@ -0,0 +1,126 @@
...
+        table = _geopandas_to_arrow(df, index=preserve_index)
+        # table = pa.Table.from_pandas(df, preserve_index=preserve_index,
+        #                              schema=schema)
```
jsignell, line 53: Don't need this right?

dask_geopandas/io/parquet.py
```diff
@@ -0,0 +1,126 @@
...
+    # pyarrow is imported here, but is an optional dependency
+    from dask.dataframe.io.parquet.arrow import ArrowEngine
+except ImportError:
+    ArrowEngine = object
+
+if TYPE_CHECKING:
+    import pyarrow
+
+
+class GeoArrowEngine(ArrowEngine):
```
jsignell, line 20: Is the idea that this file would possibly contain the fastparquet engine in the future? Should we add a stub to at least suggest people use pyarrow instead in the meantime?

jorisvandenbossche: I am not planning to add a fastparquet engine myself. To start, geopandas would need to be extended to work with fastparquet as well (we leave the geo metadata parsing to geopandas, and that is currently written with the pyarrow api's)

We should certainly clarify that you can't specify `engine="fastparquet"/"pyarrow"` at the moment, because that will select dask.dataframe's backend, which will fail for geodataframes.

dask_geopandas/io/parquet.py
jorisvandenbossche: That's also a possibility, but then I need to add `_geopandas_to_arrow = None` in the except clause (and the same for the other method), so I personally find this cleaner, but if you prefer that, can certainly change it

dask_geopandas/io/parquet.py
jorisvandenbossche: Yes, that's a good idea.

(note that those todo's first need a change in geopandas to support those in `_geopandas_to_arrow` and `_arrow_to_geopandas`)

dask_geopandas/io/parquet.py
jorisvandenbossche: Yes, will clean-up

jorisvandenbossche (commented):
Thanks for the review!

dask_geopandas/io/parquet.py
jsignell: No that makes sense. I am happy to leave it.

dask_geopandas/io/parquet.py
jsignell: Oh yeah I don't think we need to actually write a fastparquet engine. I missed the fixed engine part. I think it makes sense to raise an error if the user tries to pass an engine.

jsignell:
Thanks for doing this work!

PR #14, opened 2020-08-14
ENH: Implement basic spatial partitioning

jorisvandenbossche:
Working towards https://github.com/jsignell/dask-geopandas/issues/8

For now I added a `spatial_partitions` attribute, which holds a GeoSeries with length of `npartitions`. And I implemented the `cx` indexer to slice based on the coordinates, making use of the spatial partitioning.  
So that ensures you can do something like `ddf.cx[-180:-70, 0:-80]`, and it will based on this specified bounding box first select the appropriate partitions, and then further select the rows of those partitions. 

There are still lots of todo's / things I am uncertain about:

- We need to add some "repartition" method that shuffles based on given partition bounds or that first calculates optimal partition bounds (there is a version in the old repo: https://github.com/mrocklin/dask-geopandas/blob/8133969bf03d158f51faf85d020641e86c9a7e28/dask_geopandas/core.py#L339-L387)
- Need to ensure that the spatial partitions are propagated in methods and operations. But I am not fully sure what's the best way to do this in general (without overriding each of the inherited methods from dask)
- For some spatial operations, we will also need to update the spatial partitions (eg when doing a buffer)
- Implementing a spatial join using those spatial partitions should be relatively straightforward (https://github.com/mrocklin/dask-geopandas/blob/8133969bf03d158f51faf85d020641e86c9a7e28/dask_geopandas/core.py#L409-L430)

### DIFF
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -2,11 +2,14 @@ import numpy as np
 import pandas as pd
 
 import dask.dataframe as dd
-from dask.dataframe.core import _emulate, map_partitions, elemwise
+from dask.dataframe.core import _emulate, map_partitions, elemwise, new_dd_object
+from dask.highlevelgraph import HighLevelGraph
 from dask.utils import M, OperatorMethodMixin, derived_from, ignore_warning
+from dask.base import tokenize
 
 import geopandas
 from shapely.geometry.base import BaseGeometry
+from shapely.geometry import box
 
 
 def _set_crs(df, crs):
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
I updated this PR to ensure some basic operations already preserve the spatial partitioning information (geometry attribute, getitem, a few geometry properties). More complete coverage can then be done in follow-up PRs.

Feedback/review is certainly still very welcome! But I am going to merge this now, so I can more easily work on separate follow-up PRs, to not pile everything in a single PR.

PR #27, opened 2020-10-16
Adapt html repr to recognize GeoDataFrame

jorisvandenbossche:
This depends a bit on the exact output of Dask, but worse what can happen if dask changes is that we fall back to the original dask version as is now.


jsignell (approved):
This makes sense to me.

PR #32, opened 2021-03-05
REF: resolve numpy FutureWarning

martinfleis:
We now get numpy FutureWarning due to LineString within array.

```
FutureWarning: The input object of type 'LineString' is an array-like implementing one of the corresponding protocols (`__array__`, `__array_interface__` or `__array_struct__`); but not a sequence (or 0-D). In the future, this object will be coerced as if it was first converted using `np.array(obj)`. To retain the old behaviour, you have to either modify the type 'LineString', or assign to an empty array created with `np.empty(correct_shape, dtype=object)`.
```

Since `from_shapely` does not need an array and can work with a list, I am passing it directly. Broken tests are unrelated (were already broken).

### DIFF
--- a/dask_geopandas/backends.py
+++ b/dask_geopandas/backends.py
@@ -9,7 +9,6 @@ from dask.dataframe.utils import (
 from dask.dataframe.extensions import make_array_nonempty, make_scalar
 from dask.base import normalize_token
 
-import numpy as np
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (approved):
Thanks!

PR #33, opened 2021-03-05
BUG: fix explode moved from GeoPandasBase in geopandas

martinfleis:
Explode implementation was moved from GeoPandasBase in https://github.com/geopandas/geopandas/pull/1720. This should fix the implementation here.


jorisvandenbossche (approved):
The docstring is going to be slightly off for GeoSeries, but let's worry about that later ;)

PR #34, opened 2021-03-05
REF: remove custom write_partition in parquet io

martinfleis:
This removes the custom `write_partition` and uses the one implemented directly in dask https://github.com/dask/dask/pull/6505. 

It also fixes the CI failures.

This and #33 should make CI green again.


martinfleis:
Hmm, this is still failing for Python 3.7. It is fixed for 3.8 though (and 3.9 in my local env). Versions of all relevant libraries are the same... any idea why is that?

jorisvandenbossche:
Hmm, it seems to have lost the CRS. Not directly an idea (would need to try out locally). The same version of dask is installed in both cases.

jorisvandenbossche:
Let's already merge this since that makes it at least usable again?

martinfleis:
> Let's already merge this since that makes it at least usable again?

I am fine with that. I wasn't able to reproduce it locally with Python 3.7.

PR #36, opened 2021-03-05
CI: move CI to GHA

### LINKED ISSUE #35: CI: expand CI and move it to GHA
I think that we should expand CI to test on windows and mac and to include master versions of key packages. Since testing on windows/mac is not supported on Travis (maybe mac is if I remember?), I'd suggest to follow the geopandas model and switch to GitHub actions. 

Happy to do a PR.

martinfleis:
Closes #35

I have removed python 3.6 and added 3.9, which is run on Linux, windows and mac. Dev env uses master of geopandas, pygeos and dask.

It should also start using codecov for coverage.


martinfleis:
Oh, yeah. It won't activate until merged since there's no GHA workflow yet.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,67 @@
...
+          conda info
+          conda list
+
+      - name: Test
+        shell: bash
+        run: |
+          source activate test
+          pytest -v -r s --color=yes --cov=dask_geopandas --cov-append --cov-report term-missing --cov-report xml .
+
+          - uses: codecov/codecov-action@v1
```
jorisvandenbossche, line 67: identation is off here?

martinfleis: good catch, fixed

jorisvandenbossche (approved):

jorisvandenbossche:
Let's test it in master

martinfleis:
Well, the good news is that CI seems to work. The bad that it fails.

jorisvandenbossche:
It's strange, it's the same test that is failing as on travis, but with a different error ...

jorisvandenbossche:
But now it seems as something we should be able to solve

jorisvandenbossche:
It might be a change in pyproj. If I update pyproj to the latest 3.0.1, then I can reproduce the error.

But given it's a strange decoding error, that might be the thread-safety issue that came up recently.

jorisvandenbossche:
Very short term "solution" might be to pin pyproj to an older version in our CI to get that green again.

A more proper short term solution *might* be to ensure that each partition gets a **copy** of the CRS, while now we set the same CRS object on each geopandas geodataframe (which are accessed from different threads):

https://github.com/geopandas/dask-geopandas/blob/6fd01e502cc9356b7c9e5b7f09748d30acd15e4c/dask_geopandas/core.py#L155-L157

https://github.com/geopandas/dask-geopandas/blob/6fd01e502cc9356b7c9e5b7f09748d30acd15e4c/dask_geopandas/core.py#L15-L19

Eventually this will be fixed in pyproj as well. @snowman2 just opened a PR for that!

snowman2:
This is likely the issue: https://github.com/pyproj4/pyproj/issues/589

I would be surprised if there is an older version of pyproj that worked. The issue appears to be kind of random. 

You could test out this fix: https://github.com/pyproj4/pyproj/pull/793

jorisvandenbossche:
> You could test out this fix: pyproj4/pyproj#793

That's what I was planning. And indeed, with pyproj master I can reproduce the failure, with that branch it's fixed!

PR #38, opened 2021-03-07
CI: pin pyproj to 2.6

jorisvandenbossche:
See discussion in https://github.com/geopandas/dask-geopandas/pull/36


martinfleis:
Can we add a comment in YAML explaining the pin? To remind us to unpin it once pyproj fix is released.

jorisvandenbossche:
Yes, good idea, added a comment

martinfleis (approved):

martinfleis:
Thanks! Good catch with dask@main.

PR #46, opened 2021-05-25
FIX: Add numpy import to example notebook

DahnJ:
Hi all! 
I'm super excited about this project, I tried out the example notebook and found that it's missing a numpy import, so I added it. 

In case I'm not going about this properly, feel free to delete this PR and fix it in the proper way.

### DIFF
--- a/notebooks/basic-intro.ipynb
+++ b/notebooks/basic-intro.ipynb
@@ -13,6 +13,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "import numpy as np\n",
     "import geopandas\n",
     "\n",
     "import dask.dataframe as dd\n",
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Thanks, this is certainly a proper change! 
My one comment would be to maybe put the import more at the top of the notebook in the cell with the other imports.

(yeah, I have numpy configured to be automatically imported in my IPython profile, which is handy for myself, but then I forget to add it for public notebooks .. ;))

DahnJ:
> Thanks, this is certainly a proper change!
> My one comment would be to maybe put the import more at the top of the notebook in the cell with the other imports.
> 
> (yeah, I have numpy configured to be automatically imported in my IPython profile, which is handy for myself, but then I forget to add it for public notebooks .. ;))

Makes sense, Numpy import added to the other imports.

( Ah, that's clever! My solution to this is to have a quick bash function that creates (copies and renames) a default notebook with my usual imports and structure :))

jorisvandenbossche (approved):
Thanks!

PR #50, opened 2021-06-03
Add compatibility for Dask 2021.06.0

jrbourbeau:
This PR adds some compatibility updates for the upcoming `dask=2021.06.0` release. xref https://github.com/dask/dask/pull/7743#issuecomment-853693448

### DIFF
--- a/dask_geopandas/backends.py
+++ b/dask_geopandas/backends.py
@@ -13,14 +13,10 @@ from shapely.geometry.base import BaseGeometry
 import geopandas
 from geopandas.array import GeometryArray, GeometryDtype, from_shapely
 
-DASK_2021_05_1 = str(dask.__version__) >= LooseVersion("2021.05.1")
-DASK_2021_06_0 = str(dask.__version__) > LooseVersion("2021.05.1")
+DASK_2021_06_0 = str(dask.__version__) >= LooseVersion("2021.06.0")
 
 if DASK_2021_06_0:
     from dask.dataframe.dispatch import make_meta_dispatch
-    from dask.dataframe.backends import _nonempty_index, meta_nonempty_dataframe
-elif DASK_2021_05_1:
-    from dask.dataframe.utils import make_meta as make_meta_dispatch
     from dask.dataframe.backends import _nonempty_index, meta_nonempty_dataframe
 else:
     from dask.dataframe.core import make_meta as make_meta_dispatch
--- a/dask_geopandas/backends.py
+++ b/dask_geopandas/backends.py
@@ -13,13 +13,17 @@ from shapely.geometry.base import BaseGeometry
 import geopandas
 from geopandas.array import GeometryArray, GeometryDtype, from_shapely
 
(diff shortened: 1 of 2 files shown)


dask_geopandas/backends.py
```diff
@@ -13,13 +13,17 @@
...
-DASK_2021_06 = str(dask.__version__) > LooseVersion("2021.05.0")
+DASK_2021_05_1 = str(dask.__version__) >= LooseVersion("2021.05.1")
+DASK_2021_06_0 = str(dask.__version__) > LooseVersion("2021.05.1")
 
-if DASK_2021_06:
-    from dask.dataframe.utils import make_meta
+if DASK_2021_06_0:
+    from dask.dataframe.dispatch import make_meta_dispatch
+    from dask.dataframe.backends import _nonempty_index, meta_nonempty_dataframe
+elif DASK_2021_05_1:
```
jrbourbeau, line 22: This branch is just to cover the `dask=2021.05.1` release. I'm not sure if you'd prefer to keep this extra code branch, or just say "dask 2021.05.1 was a bad release, don't use it"

martinfleis: I would probably go with "don't use 2021.05.1".

martinfleis: Depending on when the next dask release will drop, we may need to add a pin to CI envs, which will now load 2021.05.1 and fail if we remove this condition.

jrbourbeau: > I would probably go with "don't use 2021.05.1"

+1 I'll update this PR accordingly

We're planning to release tomorrow (xref https://github.com/dask/community/issues/162). We can also just wait to merge this PR until `dask=2021.06.0` is out, which should help keep CI passing

martinfleis: > We can also just wait to merge this PR until dask=2021.06.0 is out, which should help keep CI passing

Sounds good to me. I'll keep an eye on the release and merge after. Thanks!

jrbourbeau: Alright, just updated this PR to simplify things a bit. I can ping this PR again once the next release is out 👍

jakirkham: > I would probably go with "don't use 2021.05.1".

Probably worth adding a version constraint to `setup.py`

https://github.com/geopandas/dask-geopandas/blob/dd21f7c7479b55e8af673757af3e79b1e87277f6/setup.py#L6

jrbourbeau: Just added `!=2021.05.1` to the version requirements in `setup.py`

jakirkham:
@martinfleis I think we need your help to run CI. Also think this should be ready for another look when you are ready 🙂

jorisvandenbossche:
CI is passing now except for codecov, which actually correctly points out that part of the if/else conditional import is not covered, since we only test against latest released dask. 
I opened https://github.com/geopandas/dask-geopandas/pull/53 to potentially cover that.

jorisvandenbossche:
And v0.1.0a4 was released with this

PR #51, opened 2021-06-03
Check for `crs` before allowing `set_crs`

### LINKED ISSUE #3: Check for `crs` before allowing `set_crs`
For `set_crs`, we probably want to check that the `crs` is None (not already set), and in that case raise a warning or error (maybe with a keyword to force it).

Otherwise I think people will accidentally call this assuming their geometries get transformed.

(we should also align with a similar method in GeoPandas)

Copied from: https://github.com/jsignell/dask-geopandas/pull/1/files/2953fe474021e53a364a58bc6f660c462a2d3c98#r394216055

DahnJ:
The behaviour should follow GeoPandas' behaviour, that is
- Raises ValueError if setting a different crs with `allow_override=False`
- Overwrites the crs if `allow_override=True`
- Overwrites the crs when using `crs.setter`

closes #3

### DIFF
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -14,8 +14,7 @@ from shapely.geometry import box
 
 def _set_crs(df, crs, allow_override):
     """Return a new object with crs set to ``crs``"""
-    df = df.copy(deep=False).set_crs(crs, allow_override=allow_override)
-    return df
+    return df.set_crs(crs, allow_override=allow_override)
 
 
 def _finalize(results):
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -12,10 +12,9 @@ from shapely.geometry.base import BaseGeometry
 from shapely.geometry import box
 
 
-def _set_crs(df, crs):
+def _set_crs(df, crs, allow_override):
     """Return a new object with crs set to ``crs``"""
-    df = df.copy(deep=False)
-    df.crs = crs
+    df = df.copy(deep=False).set_crs(crs, allow_override=allow_override)
     return df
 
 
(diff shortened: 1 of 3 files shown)


dask_geopandas/core.py
```diff
@@ -12,10 +12,9 @@
 from shapely.geometry import box
 
 
-def _set_crs(df, crs):
+def _set_crs(df, crs, allow_override):
     """Return a new object with crs set to ``crs``"""
-    df = df.copy(deep=False)
-    df.crs = crs
+    df = df.copy(deep=False).set_crs(crs, allow_override=allow_override)
```
martinfleis, line 17: Since `set_crs` doesn't work in place by default, do we need `copy(deep=False)` here?

tests/test_core.py
```diff
@@ -144,7 +144,10 @@ def test_geodataframe_crs(geodf_points_crs):
     assert dask_obj.partitions[1].crs == original
 
     new_crs = "epsg:4316"
-    new = dask_obj.set_crs("epsg:4316")
+    with pytest.raises(ValueError):
```
martinfleis, line 147: Can we add a 'match=` here? To ensure that we capture the correct error. Same below.

martinfleis (commented):
Thanks! This is neat as it leaves geopandas to deal with CRS in practice. I just left a few minor comments in the code.

dask_geopandas/core.py
DahnJ: I don't think we do, I removed it

tests/test_core.py
DahnJ: Added a match with `.*already has a CRS which is not equal to the passed CRS.*`

DahnJ:
> Thanks! This is neat as it leaves geopandas to deal with CRS in practice. I just left a few minor comments in the code.
> 
> edit: check also CI, linter is not happy

Ah, right, my IDE is set to a higher line length limit, so I didn't notice those. Will fix.

jorisvandenbossche (approved):
Thanks a lot! Looks good to me.

The failure on the dev environment is probably related to https://github.com/geopandas/dask-geopandas/pull/50 ?

martinfleis (approved):

martinfleis:
Thanks @DahnJ!

> The failure on the dev environment is probably related to #50 ?

Yes, breaking changes on dask master.

PR #53, opened 2021-06-07
CI: test against different dask versions + update minimum dask version

jorisvandenbossche:
@martinfleis @jsignell I don't know to what extent we want to test against different dask version (except latest release + master) in the current alpha phase of the project, but https://github.com/geopandas/dask-geopandas/pull/50 is adding some custom code for versions older than 2021.06, which would be not covered in the current CI setup

### DIFF
--- a/continuous_integration/envs/37-latest.yaml
+++ b/continuous_integration/envs/37-latest.yaml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   # required dependencies
   - python=3.7
-  - dask=2.2
+  - dask=2.18
   - distributed
   - geopandas
   - pygeos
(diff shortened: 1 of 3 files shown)


jrbourbeau (commented):
Nice! Yeah, since there's a minimum version of `dask` specified in `setup.py`, it's probably worth having one CI build pinned to that minimum `dask` version

jorisvandenbossche:
Dask 2.2 is giving failures, but is actually also quite old. I don't think we need to try to support a version of 2 years old for an initial version of dask-geopandas

jorisvandenbossche:
OK, I have updated the minimum version to 2.18, still quite old but tests pass. We can discuss the exact minimum version we want later, but going to merge this now so https://github.com/geopandas/dask-geopandas/pull/50 can be updated and merged

jsignell:
I am late to this, but I think 2.18 is plenty old enough for dask-geopandas to support.

PR #54, opened 2021-06-10
ENH: basic spatial join

jorisvandenbossche:
Updated version of the original implementation at https://github.com/mrocklin/dask-geopandas/blob/8133969bf03d158f51faf85d020641e86c9a7e28/dask_geopandas/core.py#L409-L430 written by @mrocklin

Still needs to have tests and docs etc, but I thought this could be helpful for discussion to already put up a simple spatial join algorithm that can use spatial partitioning information if available (and if not combines every partition of the left with every partition of the right dataframe).

### DIFF
--- a/dask_geopandas/sjoin.py
+++ b/dask_geopandas/sjoin.py
@@ -1,0 +1,55 @@
+import numpy as np
+import geopandas
+
+from dask.base import tokenize
+from dask.highlevelgraph import HighLevelGraph
+
+from .core import from_geopandas, GeoDataFrame
+
+
+def sjoin(left, right, how="inner", op="intersects"):
+    """
+    Spatial join of two GeoDataFrames.
+
+    ...
+    """
+    if isinstance(left, geopandas.GeoDataFrame):
+        left = from_geopandas(left, npartitions=1)
+    if isinstance(right, geopandas.GeoDataFrame):
+        right = from_geopandas(right, npartitions=1)
+
+    name = "sjoin-" + tokenize(left, right, how, op)
+    meta = geopandas.sjoin(left._meta, right._meta, how=how, op=op)
+
+    if left.spatial_partitions is not None and right.spatial_partitions is not None:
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
cc @tastatham I cleaned up my notebook and put it here: https://gist.github.com/jorisvandenbossche/20d3bd91b2d0db2522148e5c2c84d206 
Now, after cleaning it up, for some reason it started to segfault much more frequently (when drafting the notebook I only very occasionally ran into that). I created a reproducer for the crash and opened an issue at https://github.com/pygeos/pygeos/issues/361

jorisvandenbossche:
I have a patch for pygeos to avoid the segfault (https://github.com/pygeos/pygeos/pull/362)

dask_geopandas/sjoin.py
```diff
@@ -0,0 +1,62 @@
...
+    new_spatial_partitions = []
+    for i, (l, r) in enumerate(zip(parts_left, parts_right)):
+        dsk[(name, i)] = (geopandas.sjoin, (left._name, l), (right._name, r), how, op)
+        # TODO preserve spatial partitions of the output if only left has spatial
+        # partitions
+        if using_spatial_partitions:
+            lr = left.spatial_partitions.iloc[l]
+            rr = right.spatial_partitions.iloc[r]
+            # extent = lr.intersection(rr).buffer(buffer).intersection(lr.union(rr))
+            extent = lr.intersection(rr)
```
jorisvandenbossche, line 55: @mrocklin I know it's a very long time ago ;) But do you remember what the rationale was for intersecting with the union of both regions? (the `.intersection(lr.union(rr))`)

martinfleis (approved):
Looks nice. Can we open an issue to track the list of remaining things to do? Like different `how` options for example.

jorisvandenbossche:
Opened https://github.com/geopandas/dask-geopandas/issues/72

Issue #55, opened 2021-06-12
Copy `spatial_partitions` when copying `GeoDataFrame`


DahnJ:
I would expect the `spatial_partitions` to be copied when using `df.copy()`.

```python
import geopandas
import dask
import dask_geopandas
dask.config.set(scheduler='single-threaded')

df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
ddf = dask_geopandas.from_geopandas(df, npartitions=4)
ddf.calculate_spatial_partitions()
ddfc = ddf.copy()
ddfc.spatial_partitions  # is None
```

Is there a reason to not copy spatial partitions that I didn't think of?

I thought about how this could work and didn't think of anything better than overriding dask's `copy` method. Something like

```python
def copy(self):
    self_copy = super().copy()
    self_copy.spatial_partitions = self.spatial_partitions
    return self_copy
```

It would be great if `spatial_partitions` could get copied automatically, as `crs` for example does. However, `crs` is actually a property of `_meta`, which does get automatically copied by dask. Not sure how spatial partitions could be attached to the `_meta`, or if there's another mechanism that could act similarly.

If I'm headed in the right direction, then I'm happy to unit test & PR.

jorisvandenbossche:
Yes, I agree that would certainly be expected to be propagated to the copy.

I suspect there will be other methods as well where we need to fix this.

On the short term, I think your proposed fix of overriding the method like
that is fine (maybe just copy the actual spatial_partitions as well). PR
certainly welcome!
Long term it would be good to have a nicer way to have the spatial
paritions copied over, but probably something to discuss with the dask team
what this could look like.

Op za 12 jun. 2021 12:10 schreef Dahn ***@***.***>:

martinfleis:
Closed by #56

PR #56, opened 2021-06-15
Copy `spatial_partitions` when copying `GeoDataFrame`

DahnJ:
Creates a `copy()` method that copies `spatial_partitions`, as discussed in https://github.com/geopandas/dask-geopandas/issues/55

### DIFF
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -14,8 +14,7 @@ from shapely.geometry import box
 
 def _set_crs(df, crs, allow_override):
     """Return a new object with crs set to ``crs``"""
-    df = df.copy(deep=False).set_crs(crs, allow_override=allow_override)
-    return df
+    return df.set_crs(crs, allow_override=allow_override)
 
 
 def _finalize(results):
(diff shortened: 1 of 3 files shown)


jorisvandenbossche (approved):
Thanks, looks good!

jorisvandenbossche:
@DahnJ BTW, looking at the git history in this PR, you probably want to "clean up" your local master branch (it seems to include commits / merges from earlier PRs). 
For this PR it doesn't matter much (since I squashed all commits while merging anyway), but it will probably make your development / next PRs a bit easier.

DahnJ:
> @DahnJ BTW, looking at the git history in this PR, you probably want to "clean up" your local master branch (it seems to include commits / merges from earlier PRs).

Yes, you're right, I'm not sure how to "clean up" though. What operation should I perform?

jorisvandenbossche:
If there are no other commits on your master branch that you still want to keep / make a PR from, then you could do a full "reset" of your local master branch with upstream master (assuming this repo is called `upstream`, and your fork `origin`):

```
git fetch upstream
git checkout master
git reset --hard upstream/master
```

After that, it's best to always ensure to commit your work in a new branch, and not in the master branch. And then you can keep your local master branch updated with upstream's master.

PR #65, opened 2021-06-20
ENH: Implement geopandas.clip

### LINKED ISSUE #57: ENH: Implement `geopandas.clip`
I needed [`geopandas.clip`](https://geopandas.org/docs/reference/api/geopandas.clip.html) functionality the other day and I utilized `map_partitions` to that end successfully.

Reading https://github.com/geopandas/dask-geopandas/pull/54 I learned how one might be able to perform `clip` in a "smarter" way, so I gave it a try.

```python
import numpy as np
import pandas as pd
import geopandas as gpd
from dask.highlevelgraph import HighLevelGraph
import dask_geopandas

# Create data
df = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
ddf = dask_geopandas.from_geopandas(pd.concat([df]*5), npartitions=40)
mask = df.loc[df['iso_a3'].eq('CZE')]
mask['geometry'] = mask.geometry.buffer(1)

# Create more localized spatial partitions
ddf = ddf.set_crs('epsg:4326').set_index('continent').reset_index().persist()
ddf.calculate_spatial_partitions()
ddf.spatial_partitions.set_crs('epsg:4326', inplace=True) 
# ddf.spatial_partitions.plot(color='none', edgecolor='black') 

# Naive version (~2.75s on my machine)
result_naive = ddf.map_partitions(lambda x: gpd.clip(x, mask)).compute()

# Smarter version (~1.37s on my machine)
new_spatial_partitions = gpd.clip(ddf.spatial_partitions.to_frame('geometry'), mask)
intersecting_partitions = np.asarray(new_spatial_partitions.index)

name = 'clip-test'
dsk = {(name, i): (gpd.clip, (ddf._name, l), mask) for i, l in enumerate(intersecting_partitions)}
divisions = [None] * (len(dsk) + 1)
graph = HighLevelGraph.from_collections(name, dsk, dependencies=[ddf])
result = dask_geopandas.core.GeoDataFrame(graph, name, ddf._meta, tuple(divisions))
result.spatial_partitions = new_spatial_partitions

result_smart = result.compute()

pd.testing.assert_frame_equal(result_naive, result_smart)
```

Does this look like a reasonable approach? If so, then I'm happy to test & PR.

DahnJ:
A couple of notes/questions
- Is the token correct like this? (not using `gdf._name`, starting with `clip`)
- I have imported a pytest fixture from `test_core` and had to ignore flake warnings because of it. Not sure this is best way to go about it.

Commit message:
```
If no spatial partitions are available, just map_partitions
If spatial partitions are available, construct the graph
only for spatial partitions intersecting the mask.

keep_geom_type is always false for clipping the 
spatial partitions: otherwise we'd be falsely creating
new partition(s)
```

closes #57

### DIFF
--- a/dask_geopandas/clip.py
+++ b/dask_geopandas/clip.py
@@ -3,15 +3,15 @@ import geopandas
 
 from dask.highlevelgraph import HighLevelGraph
 from dask.utils import derived_from
-from dask.base import tokenize, DaskMethodsMixin
+from dask.base import tokenize
 
-from .core import GeoDataFrame
+from .core import GeoDataFrame, GeoSeries
 
 
 @derived_from(geopandas.tools)
 def clip(gdf, mask, keep_geom_type=False):
-    if isinstance(mask, DaskMethodsMixin):
-        raise NotImplementedError("Mask cannot be a Dask object.")
+    if isinstance(mask, GeoDataFrame) or isinstance(mask, GeoSeries):
+        raise NotImplementedError("Mask cannot be a Dask GeoDataFrame or GeoSeries.")
 
     if gdf.spatial_partitions is None:
         return gdf.map_partitions(
--- a/dask_geopandas/clip.py
+++ b/dask_geopandas/clip.py
@@ -3,13 +3,16 @@ import geopandas
 
 from dask.highlevelgraph import HighLevelGraph
 from dask.utils import derived_from
-from dask.base import tokenize
+from dask.base import tokenize, DaskMethodsMixin
 
 from .core import GeoDataFrame
 
 
 @derived_from(geopandas.tools)
 def clip(gdf, mask, keep_geom_type=False):
+    if isinstance(mask, DaskMethodsMixin):
+        raise NotImplementedError("Mask cannot be a Dask object.")
+
     if gdf.spatial_partitions is None:
         return gdf.map_partitions(
             lambda partition: geopandas.clip(
@@ -22,6 +25,9 @@ def clip(gdf, mask, keep_geom_type=False):
     new_spatial_partitions = geopandas.clip(
         gdf=gdf.spatial_partitions.to_frame("geometry"),
         mask=mask,
+        # keep_geom_type is always false for clipping the spatial partitions
+        # otherwise we'd be falsely creating new partition(s)
+        keep_geom_type=False,
     )
     intersecting_partitions = np.asarray(new_spatial_partitions.index)
 
--- a/dask_geopandas/clip.py
+++ b/dask_geopandas/clip.py
@@ -10,7 +10,7 @@ from .core import GeoDataFrame, GeoSeries
 
 @derived_from(geopandas.tools)
 def clip(gdf, mask, keep_geom_type=False):
-    if isinstance(mask, GeoDataFrame) or isinstance(mask, GeoSeries):
+    if isinstance(mask, (GeoDataFrame, GeoSeries)):
         raise NotImplementedError("Mask cannot be a Dask GeoDataFrame or GeoSeries.")
 
     if gdf.spatial_partitions is None:
--- a/dask_geopandas/clip.py
+++ b/dask_geopandas/clip.py
@@ -1,0 +1,38 @@
+import numpy as np
+import geopandas
+
+from dask.highlevelgraph import HighLevelGraph
+from dask.utils import derived_from
+from dask.base import tokenize
+
+from .core import GeoDataFrame
+
+
+@derived_from(geopandas.tools)
+def clip(gdf, mask, keep_geom_type=False):
+    if gdf.spatial_partitions is None:
+        return gdf.map_partitions(
+            lambda partition: geopandas.clip(
+                gdf=partition, mask=mask, keep_geom_type=keep_geom_type
+            ),
+            token="clip",
+            meta=gdf._meta,
+        )
+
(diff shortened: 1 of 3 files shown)


dask_geopandas/clip.py
```diff
@@ -0,0 +1,38 @@
...
+            lambda partition: geopandas.clip(
+                gdf=partition, mask=mask, keep_geom_type=keep_geom_type
+            ),
+            token="clip",
+            meta=gdf._meta,
+        )
+
+    new_spatial_partitions = geopandas.clip(
+        gdf=gdf.spatial_partitions.to_frame("geometry"),
+        mask=mask,
```
jorisvandenbossche, line 24: Can you add the comment about why `keep_geom_type` is always set to False for the spatial partitions here in the code as well (the one you posted in the PR top post for the commit message).  
I would maybe also add `keep_geom_type=False` in the line above to be explicit in the code.

dask_geopandas/clip.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+    intersecting_partitions = np.asarray(new_spatial_partitions.index)
```
jorisvandenbossche, line 26: I assume this will only work if you have a default integer index. Now, this if for the spatial_partitions (not the actual gdf), so there we can actually expect/enforce this. 
Not for this PR, but we should probably add a check when creating the GeoDataFrame that the `spatial_partitions` has a default index.

jorisvandenbossche: -> opened https://github.com/geopandas/dask-geopandas/issues/66 about this

dask_geopandas/clip.py
```diff
@@ -0,0 +1,38 @@
...
+
+    name = f"clip-{tokenize(gdf, mask, keep_geom_type)}"
```
jorisvandenbossche, line 28: I think this name is fine

tests/test_clip.py
```diff
@@ -0,0 +1,23 @@
...
+import dask_geopandas
+from .test_core import geodf_points  # noqa: F401
+
+
+def test_clip(geodf_points):  # noqa: F811
+    dask_obj = dask_geopandas.from_geopandas(geodf_points, npartitions=2)
+    dask_obj.calculate_spatial_partitions()
+    mask = geodf_points.iloc[:1]
+    mask["geometry"] = mask["geometry"].buffer(2)
+    expected = dask_obj.map_partitions(lambda gdf: geopandas.clip(gdf, mask)).compute()
```
jorisvandenbossche, line 12: I would calculate the expected with geopandas in this case (`expected = geopandas.clip(geodf_points, mask)`)

jorisvandenbossche (commented):
Thanks a lot! Added a few small comments. Also: 

- Can you add a check that `mask` is not a dask object? (maybe we can support that in the future, but for now I would keep it simple)

DahnJ:
Thanks for the review, I implemented the changes

> * Can you add a check that `mask` is not a dask object? (maybe we can support that in the future, but for now I would keep it simple)

I checked `isinstance(mask, DaskMethodsMixin)`, not sure if that's the right class to check.

DahnJ:
This PR seems to be a bit stale, is there any more action required from my side?

dask_geopandas/clip.py
```diff
@@ -0,0 +1,44 @@
...
+from dask.highlevelgraph import HighLevelGraph
+from dask.utils import derived_from
+from dask.base import tokenize, DaskMethodsMixin
+
+from .core import GeoDataFrame
+
+
+@derived_from(geopandas.tools)
+def clip(gdf, mask, keep_geom_type=False):
+    if isinstance(mask, DaskMethodsMixin):
```
martinfleis, line 13: I don't think this is the correct check. See:

```py
>>> from dask.base import tokenize, DaskMethodsMixin
>>> from dask_geopandas import GeoDataFrame, GeoSeries

>>> isinstance(GeoDataFrame, DaskMethodsMixin)
False
```

I'd say checking against GeoSeries and GeoDaraFrame would be enough.

martinfleis (commented):
@DahnJ thanks for the ping. One comment on type check but apart from that it seems to be ready.

dask_geopandas/clip.py
DahnJ: Thanks for the comment and sorry for the delay. I have changed the check to `GeoSeries` or `GeoDataFrame`

dask_geopandas/clip.py
```diff
@@ -0,0 +1,44 @@
...
+from dask.highlevelgraph import HighLevelGraph
+from dask.utils import derived_from
+from dask.base import tokenize
+
+from .core import GeoDataFrame, GeoSeries
+
+
+@derived_from(geopandas.tools)
+def clip(gdf, mask, keep_geom_type=False):
+    if isinstance(mask, GeoDataFrame) or isinstance(mask, GeoSeries):
```
martinfleis, line 13:
```suggestion
    if isinstance(mask, (GeoDataFrame, GeoSeries)):
```

Shorter this way :).

martinfleis (approved):
Thanks @DahnJ!

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @DahnJ !

PR #69, opened 2021-07-06
ENH: implement parallel `dissolve`

martinfleis:
Implementing `dissolve`. 

I am experimenting with two options:

1. First dissolve geometries within each partition, then shuffle resulting (smaller) ddf (potentially expensive but necessary) to ensure that geometries belonging to the same group are within the same partition to be unioned into a single one and then does dissolve within new partitions. (currently commented out)
2. Shuffle ddf at the beginning to ensure that geometries belonging to the same group are within the same partition to be unioned into a single one and then does dissolve within new partitions.

The rationale behind the first option is that we will need to shuffle smaller ddfs than in the second one. But testing locally on my 4 cores with the default scheduler, it seems to be slower than the second. Though I haven't tried on a proper benchmark yet (will report). 

Missing:

- [x] Figure out implementation
- [x] Tests
- [x] Docstring


jorisvandenbossche:
My intuition (but which is not necessarily correct :)) is that the first approach will almost always be the better one for larger datasets. Reasons for that:

- This operation is basically a groupby (where the aggregation function for the geometry column is a union), and dask implements groupby in such a step-wise manner, and not with a shuffle (since shuffling is generally very expensive)
- Union can be an expensive operation, and I thought that doing it in steps (first union subsets and then union the results) can generally improve performance of unary_union with many polygons

martinfleis:
My intuition is the same, that is why I started with that approach first. I'll try something more complex. 

> Union can be an expensive operation, and I thought that doing it in steps (first union subsets and then union the results) can generally improve performance of unary_union with many polygons

That is not true anymore with the recent changes to overlay in GEOS, the difference is marginal now (while it was massive before) (ref https://github.com/pygeos/pygeos/issues/292).

jorisvandenbossche:
> That is not true anymore with the recent changes to overlay in GEOS, the difference is marginal now (while it was massive before) (ref pygeos/pygeos#292).

Ah yes, forgot about that issue :-)

For an initial implementation, it might make sense to follow a similar pattern as the geopandas version: two groupby calls, one for the attributes, one for the geometries. And then concat the results. I think that should be possible to implement in terms of dask dataframe calls.

(longer term it might be good to integrate with the lower level groupby implementation of dask to do it in a single step, but I don't know how straightforward it would be to extent that.

jorisvandenbossche:
(note that the manual map_partition / shuffle / map_partition workflow won't work correctly for attribute aggregations; for example a "mean" needs to be handled customly (tracking sum and length per partition) and cannot be done naively in stepwise manner, which is what is handled by dask's groupby)

martinfleis:
> note that the manual map_partition / shuffle / map_partition workflow won't work correctly for attribute aggregations

Good point.

I have added groupby-based implementation but it is currently the slowest of the three. I'll try to do some reasonable benchmarks in the following days.

martinfleis:
I realised one more difference in here. While the shuffle method returns partitioned GeoDataFrame, the groupby method returns a single partition. Since dissolve is essentially groupby, I'd say we should be consistent with dask here, hence use groupby method.

This is also the reason why is the shuffle method faster - it uses all workers most of the time while groupby does the last bit in a single worker.

martinfleis:
Dropping this as a reference here https://saturncloud.io/docs/reference/dask_groupby_aggregations/ . I will play with these options but I think I will clean this up and keep only the groupby solution as it is way more robust than the other two.

jorisvandenbossche:
Wow, that's a nice explanation / illustration!

martinfleis:
As conclusion:

- `shuffle-dissolve` option is always faster when your data fits in memory and is easy to sort and shuffle around. That is because it calls `unary_union` only once per geometry and the overhead of in-memory shuffling is much lower than the time we need to spend in pygeos.
- `groupby` also does the `unary_union` at least twice - once in a chunk and then as a final aggregation. With custom `split_every` even 3x. That is where the difference comes from.
- Since the most expensive bit of `dissolve` is the `unary_union`, it helps if you specify `split_out` parameter in aggregation and set it at least to a number of workers. That way we can parallelize even the last `unary_union`. 
- `groupby` option is much better in out-of-core context as the dask machinery behind it is super optimized to scale up. The shuffle overhead gets more pronounced as dask starts spilling the data on disk. It is also better when you have your workers spread across different nodes hence the inter-worker communication is slow and you need to send as least data as you need.

The actual timings obviously depend a lot on specific cases (complexity of geometries, the spread of groups across partitions, number of partitions...).

If you want to take your geopandas.GeoDataFrame that fits in your memory and speed-up dissolve, then `shuffle-dissolve` is a good option. Otherwise, it is not. 

I will finalise this PR with the groupby implementation and add a notebook, which will also include the `shuffle-dissolve` option in case someone wants to use it manually.

dask_geopandas/core.py
```diff
@@ -395,6 +395,41 @@ def to_parquet(self, path, *args, **kwargs):
...
+        split_out : int, default 1
+            Number of partitions of the output
+
+        **kwargs
+            keyword arguments passed to `groupby`
+
+        """
+
+        def union(block):
+            merged_geom = block.unary_union
```
brendan-ward, line 419: I was experimenting with this, to see if it would be possible to optionally use `coverage_union_all` from pygeos here if one has topologically clean geometries for a potential speedup.  It didn't work for the `world` ddf example in tests below because some of the blocks are only LineString geometries, which is not what I would expect (should be polygons).

Results of doing a `print(block)` inside `union` function.
```
0    LINESTRING (0.00000 0.00000, 0.00000 1.00000)
1    LINESTRING (1.00000 1.00000, 1.00000 2.00000)
```

This might just be a misunderstanding on my part about the internal structure of the GeoSeries in dask-geopandas, but I found this a bit confusing.

brendan-ward (commented):
Thanks for working on this, @martinfleis !  Combined with good partitions that help ensure within-partition overlaps and reduce between-partition overlaps, this should add a nice speedup across multiple cores.

A couple minor suggestions and a question.

dask_geopandas/core.py
martinfleis: That is not the contents of `block`, just a placeholder we use to register geopandas objects with dask here https://github.com/geopandas/dask-geopandas/blob/99fa1a865156649d66c9ae4fa6fe119b6be6206c/dask_geopandas/backends.py#L37-L41

It may be interesting to optionally use `coverage_union_all` but I'd like to do that in sync with GeoPandas. We should try to keep the API as similar as possible.

dask_geopandas/core.py
```diff
@@ -395,6 +395,41 @@ def to_parquet(self, path, *args, **kwargs):
...
 
+    def dissolve(self, by=None, aggfunc="first", split_out=1, **kwargs):
+        """Dissolve geometries within `groupby` into single observation.
+
+        Parameters
+        ----------
+        by : string, default None
+            Column whose values define groups to be dissolved. If None,
+            whole GeoDataFrame is considered a single group.
+        aggfunc : function or string, default "first"
```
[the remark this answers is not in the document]
martinfleis: That should be quite straightforward, we create the dict on line 425 anyway. I'll have a look at that. I think that geopandas actually support dict already.

dask_geopandas/core.py
jorisvandenbossche: > It may be interesting to optionally use coverage_union_all but I'd like to do that in sync with GeoPandas. We should try to keep the API as similar as possible.

+1 on both

---

In general, with the current `meta` approach (creating dummy data), we are going to run into problems for methods that only work for a certain geometry type (because our GeometryDType is generic, the schema / meta doesn't contain any information about this). Because calling that function on `meta` to discover the output `meta` will then always error. 
The workaround for now can be to ensure the user can manually specify a `meta` in those methods were this might come up.

brendan-ward: Thanks for the clarification, that definitely was confusing but I can see where that comes from.

`coverage_union_all` was just an experiment to see what was possible, completely agreed that should be done in sync with geopandas.

notebooks/dissolve.ipynb
```diff
@@ -0,0 +1,655 @@
...
+   ],
+   "outputs": [],
+   "metadata": {}
+  },
+  {
+   "cell_type": "markdown",
+   "source": [
+    "# Dissolve with dask-geopandas\n",
+    "\n",
+    "Dask-geopandas offers distributed version of `dissolve` method based on `dask.dataframe.groupby` method. It shares all benefits and caveats of the vanilla `groupby` with a few minor differences.\n",
```
jorisvandenbossche, line 18:
```suggestion
    "Dask-geopandas offers a distributed version of the `dissolve` method based on `dask.dataframe.groupby` method. It shares all benefits and caveats of the vanilla `groupby` with a few minor differences.\n",
```

notebooks/dissolve.ipynb
```diff
@@ -0,0 +1,655 @@
...
+    "\n",
+    "Geometry column is aggregated using `unary_union` operation, which tends to be relatively costly. Therefore it is beneficial to try to use all workers in every step of the operation. GroupBy by default returns a result as a single partition, which in reality means that the final aggregation step is done within a single thread. For geometries, that means that the operation needs to _loop_ through groups of geometries coming from other partitions and call `unary_union` on every group one by one.\n",
```
jorisvandenbossche, line 20:
```suggestion
    "The geometry column is aggregated using the `unary_union` operation, which tends to be relatively costly. Therefore it is beneficial to try to use all workers in every step of the operation. GroupBy by default returns a result as a single partition, which in reality means that the final aggregation step is done within a single thread. For geometries, that means that the operation needs to _loop_ through groups of geometries coming from other partitions and call `unary_union` on every group one by one.\n",
```

notebooks/dissolve.ipynb
```diff
@@ -0,0 +1,655 @@
...
+    "    \"\"\"Shuffle and map partition\"\"\"\n",
+    "\n",
+    "    meta = ddf._meta.dissolve(by=by, as_index=False, **kwargs)\n",
+    "\n",
+    "    shuffled = ddf.shuffle(\n",
+    "        by, npartitions=ddf.npartitions, shuffle=\"tasks\", ignore_index=True\n",
+    "    )\n",
+    "\n",
+    "    return shuffled.map_partitions(\n",
+    "        ddf._partition_type.dissolve, by=by, as_index=False, meta=meta, **kwargs\n",
```
jorisvandenbossche, line 499:
```suggestion
    "        geopandas.GeoDataFrame.dissolve, by=by, as_index=False, meta=meta, **kwargs\n",
```

? 

(although `_partition_type` might be a quite stable private attribute)

jorisvandenbossche (commented):
Looks good!

martinfleis:
This actually doesn't work if the `aggfunc`  is not supported for all columns. E.g. using `"mean"` will raise when there is a string column. Pandas/dask ignore those columns in groupby if you pass a single function.

I am not sure how to resolve this in a smart way though.

martinfleis:
> One option would be to mirror the dual groupby we have in geopandas

That doesn't help, in fact. Because this fails in dask (while pandas is fine with it and ignores invalid columns)

```py
ddf.drop(columns="geometry").groupby(by).aggregate("mean")
```

I think that for `dask-geopandas`, it may be best to force users to select columns first if they want to compute mean while having string columns in dataframe. I am not sure how else to do dissolve without rewriting half of groupby. This means we can go with the implementation in this PR (but it also means that porting this back to geopandas is not feasible).

jorisvandenbossche:
Ah, yes, the issue of automatically dropping columns ... 

> I think that for `dask-geopandas`, it may be best to force users to select columns first if they want to compute mean while having string columns in dataframe

I think that's fine to be a bit more strict about this in dask-geopandas (especially since dask is already doing the same, right?). In general dask wants to know in advance what the result will look like (the structure / dtypes of the result), which is something pandas is sometimes more flexible about. 

Maybe just add a note about it in the docstring?

martinfleis:
> especially since dask is already doing the same, right?

In `aggregate` yes. But this will work

```py
ddf.groupby("foo").mean()
```

This won't
```py
ddf.groupby("foo").agg("mean")
```

I'll expand the docstring and add support of `dissolve()` (i.e. `by=None`) which I forgot about.

dask_geopandas/core.py
```diff
@@ -445,6 +445,60 @@ def to_parquet(self, path, *args, **kwargs):
...
+
+        Parameters
+        ----------
+        by : string, default None
+            Column whose values define groups to be dissolved. If None,
+            whole GeoDataFrame is considered a single group.
+        aggfunc : function,  string or dict, default "first"
+            Aggregation function for manipulation of data associated
+            with each group. Passed to dask `groupby.agg` method.
+            Note that ``aggfunc`` needs to be applicable to all column (i.e. ``"mean:``
```
jorisvandenbossche, line 459:
```suggestion
            Note that ``aggfunc`` needs to be applicable to all columns (i.e. ``"mean"``
```

dask_geopandas/core.py
```diff
@@ -445,6 +445,60 @@ def to_parquet(self, path, *args, **kwargs):
...
+            cannot be used with string dtype). Select only required columns before
+            ``dissolve`` or pass a dictionary mapping ``aggfunc`` to each column
+            separately.
```
jorisvandenbossche, line 462:
```suggestion
            ``dissolve`` or pass a dictionary mapping to ``aggfunc`` to specify the
            aggregation function for each column separately.
```

dask_geopandas/core.py
```diff
@@ -445,6 +445,60 @@ def to_parquet(self, path, *args, **kwargs):
...
+        if isinstance(aggfunc, dict):
+            data_agg = aggfunc
+        else:
+            data_agg = {col: aggfunc for col in self.columns.drop(drop)}
+        data_agg[self.geometry.name] = merge_geometries
+        aggregated = self.groupby(by=by, **kwargs).agg(
+            data_agg,
+            split_out=split_out,
+        )
+        return aggregated.set_crs(self.crs)
```
jorisvandenbossche, line 500: Do we only want to set the crs of the `meta` here? Or is this actually needed for the results of dissolve in each partition? 
Because if it's only for meta, I think we should do this more targeted (eg `aggregated._meta.crs = ..`). And if not, that sounds as a bug in geopandas as well?

tests/test_core.py
```diff
@@ -436,3 +436,47 @@ def get_geometry(partition):
...
+        self.ddf = dask_geopandas.from_geopandas(self.world, npartitions=4)
+
+    def test_default(self):
+        gpd_default = self.world.dissolve("continent")
+        dd_default = self.ddf.dissolve("continent").compute()
+        assert_geodataframe_equal(gpd_default, dd_default, check_like=True)
+
+    def test_sum(self):
+        gpd_sum = self.world.dissolve("continent", aggfunc="sum")
+        dd_sum = self.ddf.dissolve("continent", aggfunc="sum").compute()
```
jorisvandenbossche, line 455: The fact that this works (and automatically drops columns) seems a bit at odds with the note you added in the docstring?

martinfleis: It doesn't drop columns, I drop them manually on line 458 before the check. It does a sum of strings so you get things like `RussiaNorwayFranceSwedenBelarusUkrainePolandAu...`.

So it is in line with the docstring above.

dask_geopandas/core.py
martinfleis: Setting CRS of the `meta` is not enough, the computed df comes without a CRS. 

> And if not, that sounds as a bug in geopandas as well?

I don't think so. I GeoPandas we also set CRS manually in 
```py
aggregated_geometry = GeoDataFrame(g, geometry=self.geometry.name, crs=self.crs)
```

This `set_crs` is doing the same manual assignment.

jorisvandenbossche: > I don't think so. I GeoPandas we also set CRS manually in

Ah, yes, I was for a moment thinking this was doing a geopandas' dissolve on each partition ... 

There is one difference though that with `set_crs` it's actually doing an extra copy of the data compared to `GeoDataFrame(..)`. But yeah, maybe something we can optimize later.

tests/test_core.py
jorisvandenbossche: Ah, yes, sum "works" for string columns ..

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #70, opened 2021-07-07
ENH: Implement Hilbert distance

tastatham:
Implementing  `hilbert_distance` for Dask-GeoPandas - based on the [`_with_hilbert_distance_column function` from SpatialPandas](https://github.com/holoviz/spatialpandas/blob/9252a7aba5f8bc7a435fffa2c31018af8d92942c/spatialpandas/dask.py#L508).

**Details**:
* Calculating the hilbert distance will allow us to partition Dask-GeoPandas using this information.
* The output only returns the calculated hilbert distance as a numpy array - not as a column
* Numba is currently supported but I have commented out these parts - we need to discuss whether we want numba as a dependency.


dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,244 @@
...
+    n = bounds.shape[1] // 2
+    # Hilbert Side len
+    side_length = 2 ** p
+    # Empty coord array
+    coords = np.zeros((bounds.shape[0], n), dtype=np.int64)
+
+    # Calculate x and y range of total bound coords - returns array
+    geom_ranges = [(total_bounds[d], total_bounds[d + n]) for d in range(n)]
+    # Calculate mid points for x and y bound coords - returns array 
+    geom_mids = [(bounds[:, d] + bounds[:, d + n]) / 2.0 for d in range(n)]
```
martinfleis, line 44: These two can be more readable without those list comprehensions or at least without `range(n)`.  We know `n` (see above) so we don't a code generalised for n dimensions here.

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,244 @@
...
+    
+    # Transform numeric to discrete int
+    for d in range(n):
```
martinfleis, line 47: Same here. `for d in range(n):` can be a bit cryptic and is not needed.

martinfleis (commented):
This looks really good and works like a charm! Thanks a lot!

> The output only returns the calculated hilbert distance as a numpy array - not as a column

That is fine I think, you can always do `ddf["hilbert"] = ddf.hilbert_distance()` if you need it as a column.

 > Numba is currently supported but I have commented out these parts - we need to discuss whether we want numba as a dependency.

The algorithm is already performant but with numba it is 5x faster (once compiled), at least with the GADM of US (3k polygons). So I'd be up for supporting numba but we can discuss the potential issues related to that tomorrow.

Can you ensure you follow the `black` & `flake8` [code styles](https://geopandas.readthedocs.io/en/latest/community/contributing.html#style-guide-linting) to make CI happy?

Regarding tests, can we add one testing the correctness of the actual Hilbert distance (including different `p`) apart from the existing check that geopandas-based and dask-geopandas-based results are equal?

dask_geopandas/core.py
```diff
@@ -307,6 +308,30 @@ def cx(self):
...
+        ----------    
+        p	: Hilbert curve parameter
+        
+        Returns
+        ----------
+        Distances for each partition
+        """
+        
+        # Compute total bounds of all partitions rather than each partition
+        total_bounds = self.total_bounds.compute()
```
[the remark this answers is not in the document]
tastatham: > I am curious if there's a way to do this lazily (not that I know one) so we don't have to call `compute` inside the method.

I realised this would raise a comment after pushing the code - It will work without calling compute. 

This was added by mistake when debugging a problem when applying the `_hilbert_distance` function to Dask gdfs. Initially I was calling total_bounds within `_hilbert_distance` , which resulted the computation of the total bounds of each partition - which created incorrect wrong.

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+    
+    if total_bounds is None:
+        total_bounds = gdf.total_bounds
+        
+    # Calculate bounds of each geom
+    bounds = gdf.bounds.to_numpy()
+    
+    # Number of dimensions
+    n = bounds.shape[1] // 2
```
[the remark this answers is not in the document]
tastatham: > I don't think this is needed. GeoPandas is planar so this will always be 2.
that's true, I will remove this line.

dask_geopandas/hilbert_distance.py
tastatham: > I don't think this is needed. GeoPandas is planar so this will always be 2.

I will change this in relation to the above

dask_geopandas/hilbert_distance.py
tastatham: > I will change this in relation to the above
again, I will change this here

dask_geopandas/core.py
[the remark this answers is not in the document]
martinfleis: Simply removing `.compute()` fails on my end (`Metadata inference failed in lambda.`) but if you can get rid of it that would be great.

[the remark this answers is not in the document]
jorisvandenbossche: Since we need to know the total bounds of all partitions before any single partition can start calculating the hilbert curve distances, I am not sure it can be avoided. 
(maybe the compute can be delayed to that dask does it later, but it will still require a full separate scan of the data).

Looking at spatialpandas, they don't call compute in [`_with_hilbert_distance_column`](https://github.com/holoviz/spatialpandas/blob/9252a7aba5f8bc7a435fffa2c31018af8d92942c/spatialpandas/dask.py#L508-L516), but that's because their [`total_bounds`](https://github.com/holoviz/spatialpandas/blob/9252a7aba5f8bc7a435fffa2c31018af8d92942c/spatialpandas/dask.py#L43-L50) attribute is not lazy, but directly returning values based on the bounds of the spatial partitions.

I am not sure we should do that as well for the `total_bounds` attribute, but at least here it would be good to check if spatial partitions are present, and if so use their total bounds (which will be much cheaper (not needing a full scan of the data) than actually calculating the full total_bounds, and be good enough for the use case here)

jorisvandenbossche:
> I totally forgot to follow the styling rules! I will update this

I can recommend to use pre-commit! (https://geopandas.readthedocs.io/en/latest/community/contributing.html#style-guide-linting)

martinfleis:
>  I did think about this but didn't know how to define the "correctness of the hilbert distance" without referencing another package

It can be indirect. You can have the expected order of geometries and test if sorting according to Hilbert distance produces the same order. I'd say that for some dummy geometries, you can also hard-code expected value (just to test it does not unexpectedly change by some future PR).

tastatham:
> > I totally forgot to follow the styling rules! I will update this
> 
> I can recommend to use pre-commit! (https://geopandas.readthedocs.io/en/latest/community/contributing.html#style-guide-linting)

Thanks! this is useful

dask_geopandas/core.py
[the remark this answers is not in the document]
tastatham: > Since we need to know the total bounds of all partitions before any single partition can start calculating the hilbert curve distances, I am not sure it can be avoided.

This ^. We can discuss this tomorrow.

> Simply removing `.compute()` fails on my end (`Metadata inference failed in lambda.`) but if you can get rid of it that would be great.

You're right, maybe I was thinking off a previous step.

dask_geopandas/hilbert_distance.py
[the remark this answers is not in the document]
tastatham: resolved with bc6fd49

dask_geopandas/hilbert_distance.py
tastatham: resolved with bc6fd49

dask_geopandas/hilbert_distance.py
tastatham: resolved with bc6fd49

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,261 @@
...
+    Returns
+    ---------
+    Array of hilbert distances for each geom
+    """
+
+    if total_bounds is None:
+        total_bounds = gdf.total_bounds
+
+    # Calculate bounds of each geom
+    bounds = gdf.bounds.to_numpy()
```
jorisvandenbossche, line 34: This is only needed for non-point geometries, I think? We should probably special case points here to just take x/y coordinates as `geom_mids`

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,261 @@
...
+    # Calculate x and y range of total bound coords - returns array
+    geom_ranges = [
+        (total_bounds[0], total_bounds[2]),
+        (total_bounds[1], total_bounds[3]),
+    ]
+    # Calculate mid points for x and y bound coords - returns array
+    geom_mids = [
+        ((bounds[:, 0] + bounds[:, 2]) / 2.0),
+        ((bounds[:, 1] + bounds[:, 3]) / 2.0),
+    ]
```
jorisvandenbossche, line 48: In general, do we know if there is an advantage to using such midpoints vs "centroid" vs "representative point" ?

(if there is not theoretical/practical reason for one or the other, we should maybe check which one is typically the cheapest to compute. EDIT: and based on a quick check, calculating the mids based on the bounds seems much faster)

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,261 @@
...
+
+    # Empty coord array
+    coords = np.zeros((bounds.shape[0], 2), dtype=np.int64)
+    # Transform continuous int to discrete int for each dimension
+    coords[:, 0] = _continuous_int_to_discrete_int(
+        geom_mids[0], geom_ranges[0], side_length
+    )
+    coords[:, 1] = _continuous_int_to_discrete_int(
+        geom_mids[1], geom_ranges[1], side_length
+    )
```
jorisvandenbossche, line 58: Personally I find it a bit easier to stack afterwards:

```
x_int = ...
y_int = ...
coords = np.stack([x_int, y_int], axis=1) 
```

jorisvandenbossche (commented):
Nice! Added a few comments.

Another idea for the tests:
- add tests using different geometry types as input (polygon, line, point)

tests/test_core.py
```diff
@@ -397,3 +397,14 @@ def test_geoseries_apply(geoseries_polygons):
 def test_geodataframe_html_repr(geodf_points):
     dask_obj = dask_geopandas.from_geopandas(geodf_points, npartitions=2)
     assert "Dask-GeoPandas GeoDataFrame" in dask_obj._repr_html_()
+
+
+def test_hilbert_distance(geodf_points):
```
jorisvandenbossche, line 402: Let's maybe start a new test file for all tests related to calculating the hilbert distance

jorisvandenbossche:
> The algorithm is already performant but with numba it is 5x faster (once compiled), at least with the GADM of US (3k polygons). So I'd be up for supporting numba but we can discuss the potential issues related to that tomorrow.

Using a point dataset (subset of the OSM GPS points, first 1 million points) and only focusing on the actual algorithm (after calculating the integer coordinates and the total bounds), I get a 200x speed-up using when using numba. 
I also quickly compared with https://github.com/PrincetonLIPS/numpy-hilbert-curve, which is faster as the plain version here, but slower as the numba version: ~1min50 without numba, ~5s for numpy-hilbert-curve and ~500ms with numba.

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,254 @@
...
+
+# @ngjit
+def _continuous_int_to_discrete_int(vals, val_range, n):
+
+    """
+    Convert an array of values from continuous data coordinates to discrete
+    int coordinates
+
+    Based on: https://github.com/holoviz/spatialpandas/blob/
+    9252a7aba5f8bc7a435fffa2c31018af8d92942c/spatialpandas/utils.py#L9
```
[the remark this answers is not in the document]
tastatham: Yes, that makes sense - I'll update this.

dask_geopandas/hilbert_distance.py
tastatham: That's true, I'll modify this function to check whether the geoms are points or not

dask_geopandas/hilbert_distance.py
tastatham: My prior was that this approach would be faster.

tests/test_core.py
tastatham: ok

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,261 @@
...
+    x_width = val_range[1] - val_range[0]
+    res = ((vals - val_range[0]) * (n / x_width)).astype(np.int64)
+
+    # clip
+    res[res < 0] = 0
+    res[res > n - 1] = n - 1
+    return res
+
+
+def _distances_from_coordinates(p, coords):
```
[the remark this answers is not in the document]
tastatham: yes, I'll add it back in.

dask_geopandas/core.py
```diff
@@ -306,6 +307,30 @@ def explode(self):
...
+        Returns
+        ----------
+        Distances for each partition
+        """
+
+        # Compute total bounds of all partitions rather than each partition
+        total_bounds = self.total_bounds.compute()
+        # Calculate hilbert distances for each partition
+        distances = self.map_partitions(
+            lambda s: _hilbert_distance(s, total_bounds=total_bounds, p=p)
```
jorisvandenbossche, line 329:
```suggestion
            lambda s, total_bounds: _hilbert_distance(s, total_bounds, p=p), total_bounds
```

Untested, but something like this ;)

dask_geopandas/hilbert_distance.py
brendan-ward: This result makes sense: the operations are in increasing levels of complexity.  Midpoint is simple math based on bounds, centroid is more complex based on calculating the "center of mass" of the geometry (varies by geometry type), and "representative point" (point on surface) is probably yet more complex since it needs to ensure the result intersects the polygon.

More of a theoretical question (longer term): the key thing to consider here is how representative the midpoints are for ordinating geometries along the Hilbert curve: what is the tradeoff for how well the points represent the locations of the geometries for partitioning (e.g., suboptimal partitions) vs spatial operations performed against those partitions.  Put differently: using midpoint for Hilbert may produce partitions quickly, but if those are suboptimal for overlay operations and makes those much slower, then maybe it is worth a somewhat more expensive method for getting the representative points.  To do this, one would need to compare the full compute time of calculate Hilbert curve, repartition, overlay.  I'm thinking of a case like the admin boundaries of France, which includes overseas territories.  Midpoint of bounds will be far away from any of those boundaries.  Centroid will be maybe a bit better but still far away from those boundaries.  Representative point would be in continental France (I think?) and thus be more optimal for repartitioning and then overlay with other European polygons.  Though - this could easily be solved by exploding into single-part geometries...

dask_geopandas/hilbert_distance.py
brendan-ward: To carry that suggestion a bit further, handling x and y independently makes things a little easier to follow (at least for me):

```
    x_mid = ((bounds[:, 0] + bounds[:, 2]) / 2.0)
    y_mid = ((bounds[:, 1] + bounds[:, 3]) / 2.0)

    x = _continuous_int_to_discrete_int(x_mid, (total_bounds[0], total_bounds[2]), side_length)
    y = _continuous_int_to_discrete_int(y_mid, (total_bounds[1], total_bounds[3]), side_length)
    coords = np.stack([x, y], axis=1)
```

dask_geopandas/hilbert_distance.py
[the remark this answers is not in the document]
brendan-ward: The docstring does not appear to match the implementation; the implementation is for a single dimension (e.g., x) at a time, not both dimensions.  Is that correct?

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,261 @@
...
+    val_range : Ranges of x and y values ([(xmin, xmax), (ymin, ymax)])
+
+    n         : Number of discrete coords (int)
+
+    Returns
+    ---------
+    Array of discrete int coords
+    """
+
+    x_width = val_range[1] - val_range[0]
```
brendan-ward, line 90:
```suggestion
    width = val_range[1] - val_range[0]
```
(not specific to `x`)

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,261 @@
...
+    res = ((vals - val_range[0]) * (n / x_width)).astype(np.int64)
```
brendan-ward, line 91:
```suggestion
    res = ((vals - val_range[0]) * (n / width)).clip(min=0, max=n).astype(np.int64)
```
(uses ndarray `clip` instead of manual clipping below.

Note: the values will always be unsigned, which implies that you could use `np.uint32` by default - but it is based on the value of n; if `n` <= 65535, you could use `uint16`, etc.

brendan-ward (commented):
I didn't have a chance to fully review, but added a few comments to consider.

dask_geopandas/core.py
jorisvandenbossche: This indeed works more or less like that. See the docs for `map_partitions`: https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.Series.map_partitions

If we are not directly passing the arguments to the function, we also don't need a lambda:

```python
        distances = self.map_partitions(
            _hilbert_distance, total_bounds=total_bounds, p=p,
        )
```

Quickly tried the above, and then it gives a problem with the meta inference (apparently for inferring the result dtype based on the dask dataframe "meta" object, it does pass the delayed total_bounds, which then doesn't work). So we will need to manually specify the result dtype (and can then also directly specify a name for the Series): `            meta=("hilbert_distance", "int64")` passed as extra argument to `map_partitions`.


I tried those fixes with a small example, and visualized the graph:

```
df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_cities"))
ddf = dask_geopandas.from_geopandas(df, npartitions=4)
d = ddf.geometry.hilbert_distance()
d.visualize()
```

image

where you can see nicely that it first computes the total_bounds from all partitions, and passing that to the hilbert_distance task (so in practice this will require loading the data twice if they don't all fit into memory, but so not now at least those two steps are fully lazy in the task graph instead of calling compute() for the total_bounds up front)

dask_geopandas/core.py
```diff
@@ -306,6 +307,30 @@ def explode(self):
...
+        ----------
+        p	: Hilbert curve parameter
+
+        Returns
+        ----------
+        Distances for each partition
+        """
+
+        # Compute total bounds of all partitions rather than each partition
+        total_bounds = self.total_bounds.compute()
```
jorisvandenbossche, line 326:
```suggestion
        total_bounds = self.total_bounds
```

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,261 @@
...
+        geom_mids[0], geom_ranges[0], side_length
+    )
+    coords[:, 1] = _continuous_int_to_discrete_int(
+        geom_mids[1], geom_ranges[1], side_length
+    )
+
+    # Calculate hilbert distance
+    hilbert_distances = _distances_from_coordinates(p, coords)
+
+    return hilbert_distances
```
jorisvandenbossche, line 63: Currently, this is returning an array. We should maybe wrap this again in a pandas Series (preserving the indexing of the input dataframe/series), so that the result of `dask_geopandas.GeoSeries(..).hilbert_distance()` is again a Series. 
(and we should also add a test to verify it's a Series)

dask_geopandas/hilbert_distance.py
jorisvandenbossche: (yeah, in general the code still has some (now unnecessary) generalizations for more than 2 dimensions, which makes it harder to read for the simple case of 2 dimensions)

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,261 @@
+import geopandas
+import numpy as np
+
+# from numba import jit # optional - do we want numba as a dependency for dask-geopandas?
+# ngjit = jit(nopython=True, nogil=True)
```
jorisvandenbossche, line 5: As discussed on the meeting yesterday: let's simply use numba for now (and we can update later to make it imported lazily / make it an optional dependency). 

For that, you will also need to add numba to the dependencies in the different environment yaml files at https://github.com/geopandas/dask-geopandas/tree/master/continuous_integration/envs for the CI tests

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,261 @@
...
+    """
+    Calculate the hilbert distance for a GeoDataFrame based on the mid-point of
+    the bounds for each geom and total bounds of the collection of geoms
+
+    Based on: https://github.com/holoviz/spatialpandas/blob/
+    9252a7aba5f8bc7a435fffa2c31018af8d92942c/spatialpandas/dask.py#L172
+
+    Parameters
+    ----------
+    gdf          : GeoDataFrame
```
jorisvandenbossche, line 19:
```suggestion
    gdf : GeoDataFrame
```

Only a single space before (and after) the colon, see https://numpydoc.readthedocs.io/en/latest/format.html#parameters (that's the formatting issue Stefanie mentioned yesterday)

(and the same for the other docstrings)

dask_geopandas/hilbert_distance.py
jorisvandenbossche: Midpoints (and the same for centroids / representative points) might also give suboptimal results if you have a mix of large and small polygons. I was thinking it could be something to explore (later) if you could calculate the hilbert distance for eg the bounding box points, and consider those 4 points per row together when deciding the partitions. But then of course it's not a simple "sort the hilbert_distance column" to determine the partitions.

dask_geopandas/hilbert_distance.py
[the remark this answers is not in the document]
tastatham: @brendan-ward 

[The original implementation](https://github.com/galtay/hilbertcurve#introduction) (which SpatialPandas is based on) computes one-dimensional distances for n-dimensional points. The SpatialPandas documentation notes their version [is specialised for 2-dimensional points with numba acceleration](https://github.com/holoviz/spatialpandas/blob/9252a7aba5f8bc7a435fffa2c31018af8d92942c/spatialpandas/spatialindex/hilbert_curve.py#L6).

dask_geopandas/core.py
tastatham: Nice! I have added this to 0fba9a8

dask_geopandas/hilbert_distance.py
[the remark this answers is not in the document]
tastatham: resolved in bc9b632

dask_geopandas/hilbert_distance.py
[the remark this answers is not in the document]
tastatham: docstring updated in both core.py and hilbert_distance.py in ce52794 and 9cdb098

dask_geopandas/core.py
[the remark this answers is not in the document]
tastatham: updated in 0fba9a8

dask_geopandas/core.py
tastatham: resolved in 0fba9a8

dask_geopandas/hilbert_distance.py
tastatham: resolved in bc9b632

dask_geopandas/hilbert_distance.py
tastatham: resolved in ce52794

dask_geopandas/hilbert_distance.py
tastatham: > (uses ndarray `clip` instead of manual clipping below.

`clip` is much cleaner but is causing problems right now [because it's not supported by numba](https://github.com/numba/numba/issues/3465#issuecomment-844893052)

> Note: the values will always be unsigned, which implies that you could use `np.uint32` by default - but it is based on the value of n; if `n` <= 65535, you could use `uint16`, etc.

I will leave this unresolved for now.

dask_geopandas/hilbert_distance.py
tastatham: resolved in 251ae60

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,261 @@
...
+    )
+
+    # Calculate hilbert distance
+    hilbert_distances = _distances_from_coordinates(p, coords)
+
+    return hilbert_distances
+
+
+# @ngjit
+def _continuous_int_to_discrete_int(vals, val_range, n):
```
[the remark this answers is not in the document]
tastatham: resolved in 251ae60

dask_geopandas/hilbert_distance.py
tastatham: resolved in 251ae60

dask_geopandas/hilbert_distance.py
brendan-ward: It is OK to clip manually, I didn't realize that `clip` wasn't yet supported in numba (was testing manually without).  This could be added as a follow-up issue that requires the latest version of numba, but seems pretty minor priority-wise.

However, the data type part of the comment may be work further exploration; it may be a way of reducing memory consumption.

dask_geopandas/hilbert_distance.py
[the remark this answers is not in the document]
brendan-ward: I think the Hilbert curve part of the implementation in SpatialPandas is indeed for 2 dimensions, but this specific function for converting a continuous array to discrete values is for a single dimension at a time.  `vals` here is `array([val_1, val_2,..., val_n])` and `val_range` is `(val_min, val_max)`.

martinfleis:
@tastatham can you give us a short summary of recent commits and an overview of what is still missing (at least tests for sure)?

tastatham:
> @tastatham can you give us a short summary of recent commits and an overview of what is still missing (at least tests for sure)?

Friday/yesterday I added;
* Numba for hilbert_curve.py - bc9b632 & 251ae60
* Split  `_hilbert_distance` function to allow numba support - 251ae60 
* Updated hilbert distance to lazily evaluate total_bounds - 0fba9a8
* Tidied up the docstring - ce52794 & 9cdb098

Today I am:
* Creating the tests. 
* Updating output of Hilbert distance as Series

This should cover most parts that are outstanding.

dask_geopandas/hilbert_distance.py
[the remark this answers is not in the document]
tastatham: ah sorry, I got mixed up - I was reading lines 66 instead of 63. Yes you're right, I will update the docstring so it matches.

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,254 @@
...
+
+    """
+    Convert an array of values from continuous data coordinates to discrete
```
tastatham, line 111:
```suggestion
    Convert a continuous one-dimensional array to discrete int
```

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,254 @@
...
+    int coordinates
```
tastatham, line 112:
```suggestion
```

dask_geopandas/hilbert_distance.py
tastatham: export result as pd.Series resolved in 91465f3

tastatham: test created for hilbert distance and assert type is pd.Series in 8737e55

tests/test_core.py
tastatham: resolved in 8737e55

martinfleis:
@tastatham you'll need to add `hilbertcurve` to CI environments.

tastatham:
> CI environments

Done & I also cleaned up the test a little

continuous_integration/envs/38-latest.yaml
```diff
@@ -5,12 +5,14 @@ dependencies:
...
   - dask=2021.05.0
+  - numba=0.53.1
   - distributed
   - geopandas
   - pygeos
   - pyproj=2.6  # unpin when pyproj 3.1 is released with multithreading fix
   # test dependencies
   - pytest
   - pytest-cov
+  - hilbert_curve=2.0.5
```
martinfleis, line 16:
```suggestion
  - hilbertcurve
```

I think that this is the correct name. Also, no need to pin the version.

tastatham: yes that's correct, my mistake

tests/test_hilbert_curve.py
```diff
@@ -0,0 +1,28 @@
...
+from dask_geopandas import from_geopandas
+from .test_core import geoseries_points, geoseries_lines, geoseries_polygons
+
+
+@pytest.fixture(params=[geoseries_points, geoseries_lines, geoseries_polygons])
+def geom_types(request):
+    return request.getfixturevalue(request.param)
+
+
+@pytest.fixture
```
martinfleis, line 14:
```suggestion
```

This is a test, not a fixture. Pytest doesn't run it now.

tastatham: woops, that wasn't supposed to be there. fixed in c8fbd7f

continuous_integration/envs/39-dev.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
   # test dependencies
   - pytest
   - pytest-cov
-  - hilbert_curve=2.0.5
+  - hilbert_curve
```
martinfleis, line 17:
```suggestion
  - hilbertcurve
```

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,288 @@
...
+    """
+
+    # Hilbert Side len
+    side_length = 2 ** p
+
+    # Calculate x and y range of total bound coords - returns array
+    geom_ranges = [
+        (total_bounds[0], total_bounds[2]),
+        (total_bounds[1], total_bounds[3]),
+    ]
```
jorisvandenbossche, line 95:
```suggestion
    xmin, ymin, xmax, ymax = total_bounds
```

and then passing `(xmin, xmax)` to `_continuous_to_discrete` below might be a bit simpler / more explicit to read (without the need to create a temporary geom_ranges variable that contains both ranges)

tests/test_core.py
```diff
@@ -11,7 +11,7 @@
 from geopandas.testing import assert_geodataframe_equal, assert_geoseries_equal
 
 
-@pytest.fixture
+@pytest.fixture()
```
jorisvandenbossche, line 14: A small left-over from originally adding the tests here? (I think you can undo this)

jorisvandenbossche (commented):
Thanks for the updates! A few more comments

tests/test_core.py
tastatham:
```suggestion
@pytest.fixture
```
Drop unnecessary () from pytest.fixture

dask_geopandas/hilbert_distance.py
```diff
@@ -1,37 +1,89 @@
...
-    9252a7aba5f8bc7a435fffa2c31018af8d92942c/spatialpandas/dask.py#L172
+    # Compute bounds as array
+    bounds = gdf.bounds.to_numpy()
+    # Compute hilbert distances
+    distances = _hilbert_distance(total_bounds, bounds, p)
+
+    return pd.Series(distances, name="hilbert_distance")
+
+
+@ngjit
```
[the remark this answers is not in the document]
tastatham: you're right, there isn't much difference between with and without the decorator. I have merge those two functions in 98780cd

martinfleis:
@tastatham I posted this on gitter but it got lost in the following converstation. Could you make a tiny PR adding backticks around "set_geometry" (to format it as a code) in ReadMe (to get "columns of x-y points to the `set_geometry` method")? I will merge that straight away which ensures that CI workflows in this PR will run automatically. Now we have to approve it after every single commit because you're not a contributor yet. I am trying to keep an eye on that and do it as soon as possible but I'd rather not have to do that.

tastatham:
> @tastatham I posted this on gitter but it got lost in the following conversation. Could you make a tiny PR adding backticks around "set_geometry" (to format it as a code) in ReadMe (to get "columns of x-y points to the `set_geometry` method")? I will merge that straight away which ensures that CI workflows in this PR will run automatically. Now we have to approve it after every single commit because you're not a contributor yet. I am trying to keep an eye on that and do it as soon as possible but I'd rather not have to do that.

PR created #83

dask_geopandas/core.py
```diff
@@ -306,6 +307,34 @@ def explode(self):
...
 
+    def hilbert_distance(self, p=15):
+
+        """
+        A function that calculates hilbert distance for each geometry
+        in each partition of a Dask-GeoDataFrame
+
+        Parameters
+        ----------
+        p : Hilbert curve parameter
```
jorisvandenbossche, line 318: It would be good to add some explanation of what this is, how it influences the result, and some guidance on whether you should set this or not

dask_geopandas/hilbert_distance.py
brendan-ward: Data type optimization should move to a separate issue / PR so that you don't need to include it here.

dask_geopandas/hilbert_distance.py
```diff
@@ -0,0 +1,256 @@
...
+
+    Returns
+    ---------
+    One-dimensional array of discrete ints
+    """
+
+    width = val_range[1] - val_range[0]
+    res = ((vals - val_range[0]) * (n / width)).astype(np.int64)
+
+    # TO DO: When numba 0.54 releases - used.clip(res, min=0, max=n, out=res)
```
martinfleis, line 103: Can you include this in the follow-up issue?

martinfleis (commented):
I think we can essentially merge after some minor changes.

Before we do that, can you open an issue listing follow-up tasks? Some discussed yesterday, I've noted some in the code. Mention also the documentation notebook in there.

Can you add `numba` to requirements here? https://github.com/geopandas/dask-geopandas/blob/5984a6f5e4f11d96563c9275e982d78895ac755c/setup.py#L6-L10

Almost there, good job!

.pre-commit-config.yaml
```diff
@@ -1,12 +1,12 @@
...
   - repo: https://github.com/psf/black
-    rev: stable
+    rev: 21.6b0
     hooks:
       - id: black
         language_version: python3
 
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.7.9
+    rev: 3.9.2
```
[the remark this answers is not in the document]
tastatham: merged in 7c19310

martinfleis (approved):
I think we're ready to merge ;).

martinfleis:
Ehm, we're not. Can you make sure CI passes?

setup.py
```diff
@@ -7,6 +7,7 @@
     "geopandas",
     "dask>=2.18.0,!=2021.05.1",
     "distributed>=2.18.0,!=2021.05.1",
+    "numba"
```
martinfleis, line 10:
```suggestion
    "numba",
```

Issue #61, opened 2021-06-19
ENH: repartition based on an attribute


jorisvandenbossche:
Assuming you have an attribute in the GeoDataFrame that describes some form of spatial extent or entity (eg the country name, the zip code, district, ...), this column could also be used to repartition the dataframe spatially. 

Dask already somewhat supports this with the `set_index` and `repartition` methods. But in practice they don't necessarily do exactly what we want, and a helper function that uses those methods on the hood will probably be useful. 

In my [sjoin demo notebook](https://nbviewer.jupyter.org/urls/gist.githubusercontent.com/jorisvandenbossche/20d3bd91b2d0db2522148e5c2c84d206/raw/b15de263a6bc8c61a9f22a849973ed9da1726ad0/dask-geopandas-sjoin.ipynb#Creating-spatial-partitions), I repartitioned a GeoDataFrame based on a geohash column, i.e. a column with unique values and I wanted a new partition for each value in this column. 
In ended up doing something like:

```python
df_points_sorted = df_points.set_index("geohash1").sort_index()
n_partitions = df_points.index.nunique()
ddf_points_sorted = dask_geopandas.from_geopandas(df_points_sorted, npartitions=n_partitions)
ddf_points_sorted = ddf_points_sorted.repartition(
    divisions=('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'b', 'c', 'd', 'e', 'f', 'g', 'g')).persist()
```

So dask sorts the input on the index and determines partition divisions on that in `from_(geo)pandas` (or if you already start with a dask dataframe, `set_index` will do the same). However, that tries to create more or less equal size partitions, so to ensure to actually have one partition per unique value of the attribute, I repartitioned the result again with the partition bounds explicitly given to `repartition(..)` (this are the unique values in the "geohash1" column)

I suppose we could write a custom version (maybe based on how those methods in dask are implemented) that makes this a bit simpler (e.g. we also don't need to have the partitions sorted). 
I think the main question is: do we need two passes over the data to first determine all unique values in the column, and then the actual repartitioning in a second step?

jorisvandenbossche:
[...]

PR #71, opened 2021-07-07
ENH: Implement partitioning based on hilbert distance

tastatham:
Implements partitioning Dask-GeoPandas based on calculated hilbert distances - based on the `hilbert_distance` function in #70.

**Details**:
* Calculates the hilbert distance
* Adds column using calculated distances
* Sets index/re-shuffles data using new column

### DIFF
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -307,28 +307,28 @@ class _Frame(dd.core._Frame, OperatorMethodMixin):
     def cx(self):
         return _CoordinateIndexer(self)
 
-
     def hilbert_distance(self, p=15):
 
         """
         A function that calculates hilbert distance for each geometry
-        in each partition of a Dask-GeoDataFrame 
-        
+        in each partition of a Dask-GeoDataFrame
+
         Parameters
-        ----------    
+        ----------
         p	: Hilbert curve parameter
-        
+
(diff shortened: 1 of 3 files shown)


martinfleis:
Let's talk about this tonight. I would maybe prefer a different API, more generalisable.

Let's assume we have `hilbert_distance` and `geohash` functions based on which we can spatially repartition gdf. I would then prefer to have a single method (`spatial_shuffle`?) that can consume both.


```py
# using hilbert
ddf = ddf.spatial_shuffle('hilbert')

# using geohash
ddf = ddf.spatial_shuffle('geohash')
```

This API can be extended by other methods without the necessity to have two methods for each algorithm (esp. since the `repartition_` methods will be largely the same).

jorisvandenbossche:
I think we might also need APIs at two different levels: a high level one like the `spatial_shuffle` above that encapsulates both calculating the information (eg hilbert distance, geohash) and doing the actual shuffle, and a "lower level" repartitioning function based on pre-calculated data. For the hilbert case, this might simply be the existing dask `set_index`, but for geohash this is basically repartitioning based on a discrete attribute (-> https://github.com/geopandas/dask-geopandas/issues/61)

martinfleis:
Replaced by #104

Issue #74, opened 2021-07-09
BUG: GeoDataFrame.total_bounds should be a dask array with known shape


jorisvandenbossche:
```python
>>> df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_cities"))
>>> ddf = dask_geopandas.from_geopandas(df, npartitions=4)
>>> ddf.total_bounds
dask.array<from_pandas-09ff8835d3deb87b6565b111dcf696cd-total_bounds-agg, shape=(nan,), dtype=float64, chunksize=(nan,), chunktype=numpy.ndarray>
```

This results in a dask array with unknown shape and chunksize, while we know this is always a 1D length 4 array, so I suppose we should be able to tell dask this is the case.

jorisvandenbossche:
Another potential problem with the current `total_bounds` implementation is that it uses `self.reduction(..)` which, according to the docs (https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.DataFrame.reduction), is not meant to return arrays. So it might not be that easy to fix the above (unknown shape) with `reduction`.

martinfleis:
This is actually already filed in #4.

martinfleis:
Closed by #75

PR #75, opened 2021-07-09
FIX: set known shape on total_bounds result

martinfleis:
A potential solution to #74. I am just not sure what are the memory implications of `to_dask_array()`.

### DIFF
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -180,14 +180,27 @@ class _Frame(dd.core._Frame, OperatorMethodMixin):
     @property
     @derived_from(geopandas.base.GeoPandasBase)
     def total_bounds(self):
-        b = self.bounds.to_dask_array()
-        return da.array(
-            (
-                da.nanmin(b[:, 0]),  # minx
-                da.nanmin(b[:, 1]),  # miny
-                da.nanmax(b[:, 2]),  # maxx
-                da.nanmax(b[:, 3]),  # maxy
+        def agg(concatted):
+            return np.array(
+                (
+                    np.nanmin(concatted[0::4]),  # minx
+                    np.nanmin(concatted[1::4]),  # miny
+                    np.nanmax(concatted[2::4]),  # maxx
+                    np.nanmax(concatted[3::4]),  # maxy
+                )
             )
+
+        total_bounds = self.reduction(
+            lambda x: getattr(x, "total_bounds"),
+            token=self._name + "-total_bounds",
+            meta=self._meta.total_bounds,
+            aggregate=agg,
+        )
+        return da.Array(
+            total_bounds.dask,
+            total_bounds.name,
+            chunks=((4,),),
+            dtype=total_bounds.dtype,
         )
 
     @property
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Can you check how the graph looks like on a small example? (eg like https://github.com/geopandas/dask-geopandas/pull/70#discussion_r666740079)

I am mainly wondering that by splitting it into multiple (separate) nanmin/max calls on the dask array, we might load the data multiple times.

martinfleis:
Screenshot 2021-07-09 at 10 54 47

martinfleis:
I am really not sure I am interpreting this correctly but I guess we're not loading data multiple times?

martinfleis:
Just FYI, with this implementation, the only change to #70 required to avoid `compute()` is to remove it (cc @tastatham).

jorisvandenbossche:
Hmm, not fully sure how to interpret that graph .. :)

I tried both the existing implementation and the one in this PR on a bigger dataset (OSM gps points, 114 million points, 1218 partitions in parquet), and the current implementation is a bit faster, but especially uses less memory. The one in this PR runs into max memory and needs to spill to disk some intermediate results (memory was limited artificially with dask to 4GB). The new implementation also gives a much bigger task graph (15042 tasks vs 2613 tasks).

I don't think the problem is that it's loading the data multiple times, but I think it is trying to keep the data in memory for a longer time. For the existing implementation, it reads a partitions, calculates the total bounds, keeps those 4 values and release the data, and then go to the next partition. _Maybe_ the problem with this implementation is that it doesn't understand that it should first calculate all min/max values for one partition before going to the next (it's a bit difficult to see because either it's a huge graph and the dashboard doesn't show it, or either I reduce the dataset but then I only have a few seconds to look at which tasks are waiting/processing/released in the graph in real time ;))

jorisvandenbossche:
I think another solution to the original problem could be something like:

```
def total_bounds(self):
    ...
    total_bounds = self.reduction(..)
    return da.Array(total_bounds.dask, total_bounds.name, chunks=((4,),), dtype=total_bounds.dtype)
```

So basically rewrapping the graph in a new Array object but now with the chunk size set (you can also do `total_bounds._chunks = ((4,),)`, but that seems private API)

martinfleis:
> So basically rewrapping the graph in a new Array object

I thought you also wanted to get rid of `reduction`? But I agree that the solution in this PR is not optimal.

jorisvandenbossche:
> I thought you also wanted to get rid of `reduction`?

Yeah, the docstring doesn't mention this case as explicitly supported, but it seems to be working, so ..

martinfleis:
Alright, let's go with the simple wrap of the current result in a dask array of a known shape. If the reduction will cause problems, we can revisit it later.

This solution also works within hilbert_distance without compute(). I guess the only issue there was an unknown shape.

jorisvandenbossche:
This still has https://github.com/geopandas/dask-geopandas/issues/78 as problem (loading the full data), but let's already merge this for now.

PR #77, opened 2021-07-09
Don't prepend task token with parent object name

jorisvandenbossche:
While inspecting the task graphs, I noticed that some methods from dask_geopandas were not showing up properly. The reason is that we had for example a task graph key such as " from_pandas-539469e91d2728523cacb4ce24809e38-total_bounds-agg-5c149902dcdf5b43358d736407dca688" (from `token=self._name + "-total_bounds"`, where in this case the parent was created from a pandas dataframe) which gets abbreviated to "from_pandas" in the visualization and html repr. With this change, it would get abbreviated as "total_bounds-agg".

I looked in dask.dataframe itself, and they don't seem to prepend the token with `self._name`. For some methods they do prepend with `series-` or `dataframe-` (this is defined as `self._token_prefix`). We could also do something similar and add a "geo-" prefix. Although I think most of our methods have a quite descriptive name.

cc @jsignell

### DIFF
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -162,8 +162,7 @@ class _Frame(dd.core._Frame, OperatorMethodMixin):
         return new
 
     def to_crs(self, crs=None, epsg=None):
-        token = "to_crs"
-        return self.map_partitions(M.to_crs, crs=crs, epsg=epsg, token=token)
+        return self.map_partitions(M.to_crs, crs=crs, epsg=epsg)
 
     def copy(self):
         """Make a copy of the dataframe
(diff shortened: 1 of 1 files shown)


dask_geopandas/core.py
```diff
@@ -163,7 +162,7 @@ def set_crs(self, value, allow_override=False):
         return new
 
     def to_crs(self, crs=None, epsg=None):
-        token = f"{self._name}-to_crs"
+        token = "to_crs"
         return self.map_partitions(M.to_crs, crs=crs, epsg=epsg, token=token)
```
jsignell, line 166: I'm not sure this is entirely necessary. I would expect that using the name of the method is the default dask behavior, however if we want to be explicit that is fine with me.

jsignell (approved):
Nice!

dask_geopandas/core.py
jorisvandenbossche: Indeed not needed here!

PR #84, opened 2021-07-15
DOC: Fix docstring meta for _Frame in core.py

tastatham:
The meta in `_Frame`  expects a geopandas object, not a cudf object.


martinfleis:
Nice catch!

Issue #85, opened 2021-07-21
ENH: Follow ups for the hilbert_distance function


tastatham:
The `hilbert_distance` function will shortly be merged into Dask-GeoPandas.
There was several enhancements that were discussed as follow up items, which include:

1.To pass the `total_bounds` argument to `hilbert_distance` function if this is already pre-computed.
2. Calculating the mid-points for (Multi)Point geometries in `_continuous_to_discrete_coords` is not necessary:
> - To add a check to see whether the passed geometries are (Multi)Points
> -  Only compute the mid-points for non-(Multi)Point geometries.
3. Currently the `_continuous_to_discrete` function clips discrete integer values using base Python because `numpy.clip` is not currently supported by Numba:
> - To add `numpy.clip` when it's supported by Numba - v0.54
4. Data type optimisation should be assessed
> - Total bit-width in `_continuous_to_discrete` - to use numpy.int32 instead of numpy.int64
5. To add a notebook 

Note: Will add permalinks when merged.

martinfleis:
One more thing to add is to ensure that codecov sees the coverage of numba jitted functions. We'll probably need to run that bit again without that using env variable.

tastatham:
Another follow up is to centralise/ generalise several functions that are used by both hilbert_distance` and `morton_distance`. 
* Firstly, the `_continuous_to_discrete_coords` function is imported from `hilbert_distance.py` in `morton_distance.py` and both functions use a specific - centralise
* Secondly, the `_distances_from_coordinates function is defined in both `hilbert_distance.py` in `morton_distance.py` but are used in slightly different ways - generalise

TomAugspurger:
I didn't follow this implementation closely, but is the plan to rely on the `hilbertcurve` package long-term? If so, it needs to be added to the `install_requires`: https://github.com/geopandas/dask-geopandas/blob/99fa1a865156649d66c9ae4fa6fe119b6be6206c/setup.py#L6-L11.

martinfleis:
@TomAugspurger We don't depend on it. It used only in CI. We use the custom numbs implementation ported from spatialpandas.

PR #90, opened 2021-08-04
ENH: Implement Morton distance

tastatham:
Implement morton_distance for Dask-GeoPandas. This provides another way of spatially partitioning Dask-DataFrames using another space-filling curve. 

Details:
* Calculating the hilbert distance will allow us to partition Dask-GeoPandas using this information.
* The output is returned as a Pandas Series

When compared to the Hilbert distance, there is very little difference in compute time. However, from some basic comparisons, the Hilbert distance does appear to preserve locality better. Further work needs to be carried out to compare the two (as well as a third partitioning method; Geohash).

### DIFF
--- a/continuous_integration/envs/37-latest.yaml
+++ b/continuous_integration/envs/37-latest.yaml
@@ -5,6 +5,7 @@ dependencies:
   # required dependencies
   - python=3.7
   - dask=2.18
+  - numba
   - distributed
   - geopandas
   - pygeos
@@ -12,3 +13,4 @@ dependencies:
   # test dependencies
   - pytest
   - pytest-cov
+  - hilbertcurve
--- a/continuous_integration/envs/37-latest.yaml
+++ b/continuous_integration/envs/37-latest.yaml
@@ -14,4 +14,5 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
-  - pymorton
+  - pip:
+    - pymorton
--- a/continuous_integration/envs/37-latest.yaml
+++ b/continuous_integration/envs/37-latest.yaml
@@ -14,3 +14,4 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
+  - pymorton
--- a/continuous_integration/envs/37-latest.yaml
+++ b/continuous_integration/envs/37-latest.yaml
@@ -14,5 +14,6 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
+  - pip
   - pip:
     - pymorton
--- a/continuous_integration/envs/38-latest.yaml
+++ b/continuous_integration/envs/38-latest.yaml
@@ -14,7 +14,8 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
-  - pip:
-    - pymorton
   # optional dependencies
   - pyarrow
+  - pip
+  - pip:
+    - pymorton
--- a/continuous_integration/envs/38-latest.yaml
+++ b/continuous_integration/envs/38-latest.yaml
@@ -5,6 +5,7 @@ dependencies:
   # required dependencies
   - python=3.8
   - dask=2021.05.0
+  - numba
   - distributed
   - geopandas
   - pygeos
@@ -12,5 +13,6 @@ dependencies:
   # test dependencies
   - pytest
   - pytest-cov
+  - hilbertcurve
   # optional dependencies
   - pyarrow
--- a/continuous_integration/envs/38-latest.yaml
+++ b/continuous_integration/envs/38-latest.yaml
@@ -14,5 +14,6 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
+  - pymorton
   # optional dependencies
   - pyarrow
--- a/continuous_integration/envs/38-latest.yaml
+++ b/continuous_integration/envs/38-latest.yaml
@@ -14,6 +14,7 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
-  - pymorton
+  - pip:
+    - pymorton
   # optional dependencies
   - pyarrow
--- a/continuous_integration/envs/39-dev.yaml
+++ b/continuous_integration/envs/39-dev.yaml
@@ -15,6 +15,7 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
+  - pymorton
   # optional dependencies
   - pyarrow
   - pip
--- a/continuous_integration/envs/39-dev.yaml
+++ b/continuous_integration/envs/39-dev.yaml
@@ -19,7 +19,7 @@ dependencies:
   # optional dependencies
   - pyarrow
   - pip:
-    - pymorton
+    - git+https://github.com/trevorprater/pymorton.git@master
     - git+https://github.com/geopandas/geopandas.git@master
     - git+https://github.com/pygeos/pygeos.git@master
     - git+https://github.com/dask/dask.git@main
--- a/continuous_integration/envs/39-dev.yaml
+++ b/continuous_integration/envs/39-dev.yaml
@@ -18,7 +18,7 @@ dependencies:
   # optional dependencies
   - pyarrow
   - pip:
-    - git+https://github.com/trevorprater/pymorton.git@master
+    - pymorton
     - git+https://github.com/geopandas/geopandas.git@master
     - git+https://github.com/pygeos/pygeos.git@master
     - git+https://github.com/dask/dask.git@main
--- a/continuous_integration/envs/39-dev.yaml
+++ b/continuous_integration/envs/39-dev.yaml
@@ -18,8 +18,8 @@ dependencies:
   - pymorton
   # optional dependencies
   - pyarrow
-  - pip
   - pip:
+    - pymorton
     - git+https://github.com/geopandas/geopandas.git@master
     - git+https://github.com/pygeos/pygeos.git@master
     - git+https://github.com/dask/dask.git@main
--- a/continuous_integration/envs/39-dev.yaml
+++ b/continuous_integration/envs/39-dev.yaml
@@ -4,6 +4,7 @@ channels:
 dependencies:
   # required dependencies
   - python=3.9
+  - numba
   - distributed
   - pandas
   - shapely
@@ -13,6 +14,7 @@ dependencies:
   # test dependencies
   - pytest
   - pytest-cov
+  - hilbertcurve
   # optional dependencies
   - pyarrow
   - pip
--- a/continuous_integration/envs/39-dev.yaml
+++ b/continuous_integration/envs/39-dev.yaml
@@ -17,6 +17,7 @@ dependencies:
   - hilbertcurve
   # optional dependencies
   - pyarrow
+  - pip
   - pip:
     - pymorton
     - git+https://github.com/geopandas/geopandas.git@master
--- a/continuous_integration/envs/39-dev.yaml
+++ b/continuous_integration/envs/39-dev.yaml
@@ -15,7 +15,6 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
-  - pymorton
   # optional dependencies
   - pyarrow
   - pip:
--- a/continuous_integration/envs/39-latest.yaml
+++ b/continuous_integration/envs/39-latest.yaml
@@ -5,6 +5,7 @@ dependencies:
   # required dependencies
   - python=3.9
   - dask
+  - numba
   - distributed
   - geopandas
   - pygeos
@@ -12,5 +13,6 @@ dependencies:
   # test dependencies
   - pytest
   - pytest-cov
+  - hilbertcurve
   # optional dependencies
   - pyarrow
--- a/continuous_integration/envs/39-latest.yaml
+++ b/continuous_integration/envs/39-latest.yaml
@@ -14,7 +14,8 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
-  - pip:
-    - pymorton
   # optional dependencies
   - pyarrow
+  - pip
+  - pip:
+    - pymorton
--- a/continuous_integration/envs/39-latest.yaml
+++ b/continuous_integration/envs/39-latest.yaml
@@ -14,5 +14,6 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
+  - pymorton
   # optional dependencies
   - pyarrow
--- a/continuous_integration/envs/39-latest.yaml
+++ b/continuous_integration/envs/39-latest.yaml
@@ -14,6 +14,7 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
-  - pymorton
+  - pip:
+    - pymorton
   # optional dependencies
   - pyarrow
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -347,7 +347,17 @@ class _Frame(dd.core._Frame, OperatorMethodMixin):
     def morton_distance(self, p=15):
 
         """
-        Calculate distance of geoms along Morton curve
+        Calculate the distance of geometries along the Morton curve
+        
+        The Morton distance can be used to spatially partition Dask-GeoPandas objects,
+        by mapping two-dimensional geometries along the Morton space-filing curve.
+        
+        Each geometry is represented by the midpoint of its bounds and linked to the 
+        Morton curve. The function returns a distance from the beginning 
+        of the curve to the linked point.
+        
+        Morton distance is more performant than ``hilbert_distance`` but can result in 
+        less optimal partitioning.
 
         The Morton curve is also known as Z-order https://en.wikipedia.org/wiki/Z-order
 
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -10,6 +10,7 @@ from dask.base import tokenize
 import geopandas
 from shapely.geometry.base import BaseGeometry
 from shapely.geometry import box
+from .hilbert_distance import _hilbert_distance
 
 
 def _set_crs(df, crs, allow_override):
@@ -303,6 +304,37 @@ class _Frame(dd.core._Frame, OperatorMethodMixin):
     @property
     def cx(self):
         return _CoordinateIndexer(self)
+
+    def hilbert_distance(self, p=15):
+
+        """
+        A function that calculates the Hilbert distance between the geometry bounds
+        and total bounds of a Dask-GeoDataFrame.
+        The Hilbert distance can be used to spatially partition Dask-GeoPandas objects,
+        by mapping two dimensional geometries along the Hilbert curve.
+
+        Parameters
+        ----------
+
+        p : The number of iterations used in constructing the Hilbert curve.
+
+        Returns
+        ----------
+        Distances for each partition
+        """
+
+        # Compute total bounds of all partitions rather than each partition
+        total_bounds = self.total_bounds
+
+        # Calculate hilbert distances for each partition
+        distances = self.map_partitions(
+            _hilbert_distance,
+            total_bounds=total_bounds,
+            p=p,
+            meta=pd.Series([], name="hilbert_distance", dtype="int"),
+        )
+
+        return distances
 
 
 class GeoSeries(_Frame, dd.core.Series):
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -370,7 +370,7 @@ class _Frame(dd.core._Frame, OperatorMethodMixin):
         Returns
         ----------
         type : dask.Series
-            Series containing distances from Morton curve
+            Series containing distances along the Morton curve
         """
 
         # Compute total bounds of all partitions rather than each partition
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -349,6 +349,8 @@ class _Frame(dd.core._Frame, OperatorMethodMixin):
         """
         Calculate the distance of geometries along the Morton curve
         
+        The Morton curve is also known as Z-order https://en.wikipedia.org/wiki/Z-order.
+
         The Morton distance can be used to spatially partition Dask-GeoPandas objects,
         by mapping two-dimensional geometries along the Morton space-filing curve.
         
@@ -358,8 +360,6 @@ class _Frame(dd.core._Frame, OperatorMethodMixin):
         
         Morton distance is more performant than ``hilbert_distance`` but can result in 
         less optimal partitioning.
-
-        The Morton curve is also known as Z-order https://en.wikipedia.org/wiki/Z-order
 
         Parameters
         ----------
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -359,7 +359,7 @@ class _Frame(dd.core._Frame, OperatorMethodMixin):
 
         Returns
         ----------
-        type : pandas.Series
+        type : dask.Series
             Series containing distances from Morton curve
         """
 
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -348,17 +348,17 @@ class _Frame(dd.core._Frame, OperatorMethodMixin):
 
         """
         Calculate the distance of geometries along the Morton curve
-        
+
         The Morton curve is also known as Z-order https://en.wikipedia.org/wiki/Z-order.
 
         The Morton distance can be used to spatially partition Dask-GeoPandas objects,
         by mapping two-dimensional geometries along the Morton space-filing curve.
-        
-        Each geometry is represented by the midpoint of its bounds and linked to the 
-        Morton curve. The function returns a distance from the beginning 
+
+        Each geometry is represented by the midpoint of its bounds and linked to the
+        Morton curve. The function returns a distance from the beginning
         of the curve to the linked point.
-        
-        Morton distance is more performant than ``hilbert_distance`` but can result in 
+
+        Morton distance is more performant than ``hilbert_distance`` but can result in
         less optimal partitioning.
 
         Parameters
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -340,15 +340,20 @@ class _Frame(dd.core._Frame, OperatorMethodMixin):
     def morton_distance(self, p=15):
 
         """
+        Calculate distance of geoms along Morton curve
+
+        The Morton curve is also known as Z-order https://en.wikipedia.org/wiki/Z-order
 
         Parameters
         ----------
 
-        p : The number of iterations used in constructing the Hilbert curve.
+        p : int
+            precision of the Morton curve
 
         Returns
         ----------
-        Distances for each partition
+        type : pandas.Series
+            Series containing distances from Morton curve
         """
 
         # Compute total bounds of all partitions rather than each partition
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -2,6 +2,7 @@ import numpy as np
 import pandas as pd
 
 import dask.dataframe as dd
+import dask.array as da
 from dask.dataframe.core import _emulate, map_partitions, elemwise, new_dd_object
 from dask.highlevelgraph import HighLevelGraph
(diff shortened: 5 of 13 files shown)


continuous_integration/envs/37-latest.yaml
```diff
@@ -14,3 +14,4 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
+  - pymorton
```
martinfleis, line 17:
```suggestion
  - pip
  - pip:
    - pymorton
```

`pymorton` seems to be available only on PyPI. You'll need to do the same change in all envs.

dask_geopandas/morton_distance.py
```diff
@@ -0,0 +1,76 @@
+import numpy as np
+import pandas as pd
+from dask_geopandas.hilbert_distance import _continuous_to_discrete_coords
+
+# Based on #https://code.activestate.com/recipes/
+# 577558-interleave-bits-aka-morton-ize-aka-z-order-curve/
```
martinfleis, line 6: The original code is licensed under LGPL 2.1. Can you ensure we can use it here and under which conditions?

dask_geopandas/morton_distance.py
```diff
@@ -0,0 +1,76 @@
...
+
+    gdf : GeoDataFrame
+
+    total_bounds : Total bounds
+
+    p : number of iterations
+
+    """
+    bounds = gdf.bounds.to_numpy()
+    coords = _continuous_to_discrete_coords(total_bounds, bounds, p)
```
martinfleis, line 24: The `p` parameter is the same as the Hilbert curve has? I mean, denoting the same concept?

martinfleis (commented):
Thanks a lot! First quick pass (haven't played with it yet).

Can it be combined with numba? That loop in `_distances_from_coordinates` may benefit from that with large GeoDataFrames.

dask_geopandas/morton_distance.py
jorisvandenbossche: If it's LGPL, we can't use it here. However, the recipe states that it's itself based on other code, which I think we can probably use if we refer directly to those.

dask_geopandas/morton_distance.py
```diff
@@ -0,0 +1,76 @@
...
+
+def part1by1(n):
+
+    """Interleave bits by Binary Magic Numbers"""
+
+    n &= 0x0000FFFF
+    n = (n | (n << 8)) & 0x00FF00FF
+    n = (n | (n << 4)) & 0x0F0F0F0F
+    n = (n | (n << 2)) & 0x33333333
+    n = (n | (n << 1)) & 0x55555555
```
jorisvandenbossche, line 74: Currently you work on scalar numbers? I think those bitwise operations should normally also work vectorized on the full array?

dask_geopandas/morton_distance.py
```diff
@@ -0,0 +1,76 @@
...
+    Parameters
+    ----------
+
+    coords : Coordinates in np.array format
+    """
+
+    result = np.zeros(coords.shape[0], dtype=np.int64)
+    # For each coord calculate hilbert distance
+    for i in range(coords.shape[0]):
+        result[i] = encode_morton(coords[i][0], coords[i][1])
```
jorisvandenbossche, line 45:
```suggestion
    result = encode_morton(coords[:, 0], coords[:, 1])
```

(see comment below about this (probably) working on full numpy arrays as well)

jorisvandenbossche (commented):
Naming question: if I google "morton distance", I basically get no useful result (while "hilbert distance" does give something useful). For google it, you rather need something like "morton curve". So wondering if "morton distance" is the best name (although it's of course the "distance along the curve", I suppose? 
Now, that still results in a wiki page for which the main title is "Z-order curve", so we could also consider using "Z-order" instead of "Morton" for the naming?

jorisvandenbossche:
> When compared to the Hilbert distance, there is very little difference in compute time. 

I suppose that with using the numpy-based implementation, this will become faster. A
lso, best test the performance on somewhat larger data (eg I see in your notebook that you are timing it on the "naturalearth_lowres" data, which is tiny. If you profile that timing (eg with snakeviz), you'll see that you are mostly measuring overhead, and the actual hilbert/morton encoding only is a small part). You can also easily combine the data multiple times to artificially get bigger data (`gdf=pd.concat([gdf]*1000, axis=0)`)

dask_geopandas/morton_distance.py
jorisvandenbossche: Also, https://github.com/trevorprater/pymorton has similar code and which is MIT, so that's fine. So in general this code (the specific bit values) seems widespread, so maybe no explicit reference is needed.

continuous_integration/envs/37-latest.yaml
tastatham: updated & failed in 937bbbf (39-latest.yml) ... & 2c90399. I tried adding pymorton explicitly but not sure why it's failing?

continuous_integration/envs/39-dev.yaml
```diff
@@ -15,10 +15,11 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
+  - pymorton
```
martinfleis, line 18: Remove this one.

continuous_integration/envs/39-dev.yaml
```diff
@@ -15,10 +15,11 @@ dependencies:
...
   # optional dependencies
   - pyarrow
-  - pip
   - pip:
+    - git+https://github.com/trevorprater/pymorton.git@master
```
martinfleis, line 22: Keep `pip` in there and install `pymorton` from PyPI like in the other envs. We don't need its master here.

continuous_integration/envs/39-dev.yaml
tastatham: > Remove this one.

ah, I just forgot to remove it as a conda dependency :facepalm:

updated in 72bb05d

dask_geopandas/morton_distance.py
tastatham: I believe the following logic holds true - correct me otherwise;
Using the `_continuous_to_discrete_coords` transforms each of the coordinate pairs from continuous to integers.
The p (or order of curve) value will control the number of iterations in constructing the [Hilbert/Morton] curve. 
Then calculated distances along each curve work entirely in terms of integers, where the size of the hypercube will reflect the number of bits per dimension - controlled by p;

* The 1st order curve supports values (0|1) on both x and y axis - giving us 2 bit binary numbers of the range from 0,1,2,3.
* Then the 2nd order curve supports values (0|1|2|3) on ... - giving us 4 bit binary numbers of the range 0,1...,15 and so on...

Therefore, using the same p value (or n iterations) for calculated distances along both curves will be bound by the same size hypercube.
 
The maximum discrete distances along each order are as follows;
```python
iterations = 16 # or p 
n = 2 # two-dimension
max_discrete_distances = []
for p in range(iterations):
    max_discrete_distances.append(2 ** (n*p)  - 1 ) # -1 to start index at 0
max_discrete_distances[1:]  # from 1st to 15th order curve
```

long story short, yes it denotes the same concept.

martinfleis: Cool! Thanks for the great explanation!

tastatham:
> Naming question: if I google "morton distance", I basically get no useful result (while "hilbert distance" does give something useful). For google it, you rather need something like "morton curve". So wondering if "morton distance" is the best name (although it's of course the "distance along the curve", I suppose?
> Now, that still results in a wiki page for which the main title is "Z-order curve", so we could also consider using "Z-order" instead of "Morton" for the naming?

I did think about this briefly and thought that the naming of the `morton_distance` was appropriate. This is because;
*  We are calculating the distance along the Morton/Z-order curve. 
* To not confuse Z-order with Z-dimension
*  It's consistent with the `hilbert_distance` function (which we can also change)

I'm happy to discuss this later on in the call.

dask_geopandas/morton_distance.py
tastatham: > Currently you work on scalar numbers? I think those bitwise operations should normally also work vectorized on the full array?
Yes, I can update this later to use numpy.

tastatham:
> > When compared to the Hilbert distance, there is very little difference in compute time.
> 
> I suppose that with using the numpy-based implementation, this will become faster. A
> lso, best test the performance on somewhat larger data (eg I see in your notebook that you are timing it on the "naturalearth_lowres" data, which is tiny. If you profile that timing (eg with snakeviz), you'll see that you are mostly measuring overhead, and the actual hilbert/morton encoding only is a small part). You can also easily combine the data multiple times to artificially get bigger data (`gdf=pd.concat([gdf]*1000, axis=0)`)

Good point, I will profile this and the numpy version using the artificial dataset.

dask_geopandas/morton_distance.py
jorisvandenbossche: > The p (or order of curve) value will control the number of iterations in constructing the [Hilbert/Morton] curve.

But as far as I can see, to calculate the Morton curve, we are not using any iterations? (in contrast to Hilbert curve) 
The `encode_morton` and `part1by1` seem to do a fixed set of computations independent of `p`.

(the explanation that it controls the size of the hypercube (and thus the precision of the calculated distances) is of course still valid nonetheless)

jorisvandenbossche: (I suppose the "classical" implementation also uses loops, but the bit-shifting method is an optimized version that doesn't require loops)

dask_geopandas/core.py
```diff
@@ -336,6 +337,33 @@ def hilbert_distance(self, p=15):
 
         return distances
 
+    def morton_distance(self, p=15):
+
+        """
+
```
jorisvandenbossche, line 343: TODO: expand the docstring (quick explanation, mention "also known as", maybe link to wiki page)

dask_geopandas/core.py
```diff
@@ -336,6 +337,33 @@ def hilbert_distance(self, p=15):
...
+        Parameters
+        ----------
+
+        p : The number of iterations used in constructing the Hilbert curve.
```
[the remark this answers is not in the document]
brendan-ward: Here and below, see also the [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#parameters) standard for parameters in the docstring
```
param : type
    description
```

dask_geopandas/morton_distance.py
brendan-ward: Provided I translated to numpy correctly:
```
n = np.bitwise_and(n, 0x0000FFFF)
n = np.bitwise_and(np.bitwise_or(n, np.left_shift(n, 8)), 0x00FF00FF)
n = np.bitwise_and(np.bitwise_or(n, np.left_shift(n, 4)), 0x0F0F0F0F)
n = np.bitwise_and(np.bitwise_or(n, np.left_shift(n, 2)), 0x33333333)
n = np.bitwise_and(np.bitwise_or(n, np.left_shift(n, 1)), 0x55555555)
```

dask_geopandas/core.py
```diff
@@ -336,6 +337,33 @@ def hilbert_distance(self, p=15):
...
+
+        Returns
+        ----------
+        Distances for each partition
+        """
+
+        # Compute total bounds of all partitions rather than each partition
+        total_bounds = self.total_bounds
+
+        # Calculate hilbert distances for each partition
```
brendan-ward, line 357:
```suggestion
        # Calculate Morton distances for each partition
```

brendan-ward (changes requested):
Good progress!  The main changes are to update the docstrings.

If possible, it would be good to add the numpy version of the bitwise operations; hopefully that speeds it up significantly compared to Hilbert.

dask_geopandas/morton_distance.py
martinfleis: @brendan-ward do we actually need this? On line 44 we pass numpy arrays and I'd guess that lines 69 to 73 are translated to numpy vectorised ops in numpy.

brendan-ward: Nope, you are right, they already use the numpy bitwise operators.

dask_geopandas/morton_distance.py
```diff
@@ -0,0 +1,75 @@
+import numpy as np
+import pandas as pd
+from dask_geopandas.hilbert_distance import _continuous_to_discrete_coords
```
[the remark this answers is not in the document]
tastatham: yes I agree, both `_continuous_to_discrete_coords` and it's dependency `_continuous_to_discrete` could be centralized (outside of hilbert_distance.py) 

My initial thought was that this didn't warrant a standalone script. However, both `_hilbert_distance` and `_morton_distance` also both use `_distances_from_coordinates` in [[1]](https://github.com/tastatham/dask-geopandas/blob/a1eae6ceb8cd580f3d1f8240d22e52a850c1e5bf/dask_geopandas/hilbert_distance.py#L112)
 and [[2]](https://github.com/tastatham/dask-geopandas/blob/a1eae6ceb8cd580f3d1f8240d22e52a850c1e5bf/dask_geopandas/morton_distance.py#L31) - although this is used in slightly different ways. 

Therefore it would make sense to centralize both `_continuous_to_discrete_coords` and (a generalised version of) `_distances_from_coordinates` (to work with both `_hilbert_distance` and `_morton_distance`). Does anyone have any sugestions where to centralize these functions?

dask_geopandas/morton_distance.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    Calculate Morton distances from coordinates
+
+    Parameters
+    ----------
+
+    coords : Coordinates in np.array format
+    """
+
+    result = np.zeros(coords.shape[0], dtype=np.int64)
```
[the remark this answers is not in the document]
tastatham: thanks! I realised this was still in there after the commit. Updated in 703b5b4

dask_geopandas/core.py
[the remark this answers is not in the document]
tastatham: @brendan-ward thanks for your suggestion about formatting docstring. I have updated this in 703b5b4

dask_geopandas/core.py
tastatham: Updated in 703b5b4

continuous_integration/envs/39-dev.yaml
tastatham: resolved in 9756156

dask_geopandas/morton_distance.py
tastatham: I tried both but I found the original function to be easier to read.

dask_geopandas/morton_distance.py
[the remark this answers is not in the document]
martinfleis: Can you make a note of it in the hilbetr follow-up issue (#85) and keep it as it is in this PR? We'll clean this later (I would even say that both `morton` and `hilbert` py files should go to `repartition` folder or something like that to keep the file structure clean.

continuous_integration/envs/39-dev.yaml
```diff
@@ -17,8 +17,8 @@ dependencies:
   - hilbertcurve
   # optional dependencies
   - pyarrow
-  - pip
```
martinfleis, line 20: We should always have `pip` listed among dependencies if we use it to install some. In all envs with `pymorton`. Conda adds it automatically but complains about it.

dask_geopandas/core.py
martinfleis: @tastatham This is a top-level API function (only one in this PR) so this one needs a proper docstring since this will show up in the API docs (once they're done). We need a full docstring here. Everything else is internal. This even looks that you have removed part of the docstring from here.

dask_geopandas/core.py
```diff
@@ -336,6 +337,33 @@ def hilbert_distance(self, p=15):
...
+
+        p : The number of iterations used in constructing the Hilbert curve.
+
+        Returns
+        ----------
+        Distances for each partition
+        """
+
+        # Compute total bounds of all partitions rather than each partition
+        total_bounds = self.total_bounds
```
martinfleis, line 355: When we update hilbert to allow custom bounds, we should do the same here. Can you make a public note of it somewhere?

dask_geopandas/core.py
jorisvandenbossche: You can copy the docstring you added to `_morton_distance` here

dask_geopandas/morton_distance.py
```diff
@@ -0,0 +1,80 @@
...
+    type : int
+        Integer distances from Morton curve
+    """
+
+    return np.bitwise_or(
+        _part1by1(coords[:, 0]), np.left_shift(_part1by1(coords[:, 1]), 1)
+    )
+
+
+def _part1by1(n):
```
jorisvandenbossche, line 59: Do you have an idea where the "part1by1" name comes from? (I mean, it comes from the active state recipe, but what's the reason for it? (I find it a bit cryptic name))

dask_geopandas/morton_distance.py
```diff
@@ -0,0 +1,80 @@
...
+    n : np.array
+        X or Y coordinates
+
+    Returns
+    -------
+    n : int
+        Interleaved bits
+    """
+
+    n &= 0x0000FFFF
```
jorisvandenbossche, line 74: This line is not included in the code in http://graphics.stanford.edu/~seander/bithacks.html#InterleaveBMN (the reference you added now). Do you know what it is for? (it's eg also there in pymorton https://github.com/trevorprater/pymorton/blob/f248683c19d90e904193c58bbd03e77bc2c43768/pymorton/pymorton.py#L11)

dask_geopandas/morton_distance.py
```diff
@@ -0,0 +1,80 @@
+import numpy as np
+import pandas as pd
+from dask_geopandas.hilbert_distance import _continuous_to_discrete_coords
+
+# Based on #http://graphics.stanford.edu/~seander/bithacks.html#InterleaveBMN
```
jorisvandenbossche, line 5: I would move this comment to the function that does the bit-shifting (since it's only that part that is based on that)

dask_geopandas/core.py
tastatham: I've opened a follow up issue in #92

dask_geopandas/morton_distance.py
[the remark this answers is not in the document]
tastatham: Updated issue #85 with a [new comment](https://github.com/geopandas/dask-geopandas/issues/85#issuecomment-894667314)

dask_geopandas/morton_distance.py
tastatham: yes, it's also not used in the  [`Real-Time Collision Detection book, Chapter 7, 317`](http://www.r-5.org/files/books/computers/algo-list/realtime-3d/Christer_Ericson-Real-Time_Collision_Detection-EN.pdf

Updated in 27ca9e7

dask_geopandas/morton_distance.py
tastatham: Updated in 27ca9e7

dask_geopandas/morton_distance.py
```diff
@@ -0,0 +1,80 @@
...
+        x, y coordinate pairs based on mid-points of geoms
+
+    Returns
+    -------
+    type : int
+        Integer distances from Morton curve
+    """
+
+    return np.bitwise_or(
+        _part1by1(coords[:, 0]), np.left_shift(_part1by1(coords[:, 1]), 1)
```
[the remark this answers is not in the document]
tastatham: Updated in 27ca9e7

dask_geopandas/morton_distance.py
tastatham: I agree, the naming convention is a bit cryptic but seems to be widely adopted. See [`Real-Time Collision Detection book, Chapter 7`](http://www.r-5.org/files/books/computers/algo-list/realtime-3d/Christer_Ericson-Real-Time_Collision_Detection-EN.pdf
).

I think it comes from the fact the part1by1 function inserts one zero bit between each original data bit (part), the part1by2 function inserts two zero bits between each original data bit (part) and so on...

dask_geopandas/core.py
tastatham: updated in 2c539e0

continuous_integration/envs/39-dev.yaml
tastatham: Updated in 07fdecc

dask_geopandas/morton_distance.py
jorisvandenbossche: Looking a little bit into it, this might actually still be needed because we are using int64 for the integer coordinates. With the current value of `p=15` that you are using for testing, the coordinates never become larger than the max value of an uint16. 

But if you try to run the tests with eg `p=20`, do they still pass?

The example code in the book you reference uses `uint32` type for the input values, and says it only uses the low 16 bits of those numbers (and then spreading those out with the bit interleaving). So if I am understanding correctly, if your coordinates are 16bit, you view then as 32bit when passing to the bit interleaving code. And that line `n &= 0x0000FFFF` might ensure that it only uses the lower 16 bits? (although I still don't fully understand why it is in there, as if you have a number that is larger than what can be represented by the 16bits (>65535), it gets truncated by `n &= 0x0000FFFF` anyway and then you also get a wrong result ..
But eg GEOS also has a Morton implementation, and includes it as well: https://github.com/libgeos/geos/blob/16610691611decd2b9545ea68ef41e8b94f859e6/src/shape/fractal/MortonCode.cpp#L95-L104

continuous_integration/envs/38-latest.yaml
```diff
@@ -14,5 +14,7 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
+  - pip:
+    - pymorton
   # optional dependencies
   - pyarrow
```
martinfleis, line 20:
```suggestion
  # optional dependencies
  - pyarrow
  - pip
  - pip:
    - pymorton
```

It is better to keep pip dependencies on the bottom. Also, as I mentioned before, try to always list `pip` if there are pip installed dependencies.

continuous_integration/envs/39-latest.yaml
```diff
@@ -14,5 +14,7 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
+  - pip:
+    - pymorton
   # optional dependencies
   - pyarrow
```
martinfleis, line 20:
```suggestion
  # optional dependencies
  - pyarrow
  - pip
  - pip:
    - pymorton
```

continuous_integration/envs/37-latest.yaml
```diff
@@ -14,3 +14,5 @@ dependencies:
   - pytest
   - pytest-cov
   - hilbertcurve
+  - pip:
```
martinfleis, line 17:
```suggestion
  - pip
  - pip:
```

martinfleis (commented):
Final super tiny suggestion from me. Otherwise, this is good to go. Nice job @tastatham!

dask_geopandas/morton_distance.py
tastatham: yes, that's correct. These examples assume the input values are of type `uint32`. Currently `uint64` coords are passed to the `_part1by1` function. This comes from the `_continuous_to_discrete_coords` function. 

In #70 and #90 we discussed several follow up issues; 
1. Setting the dtype for the `_continuous_to_discrete_coords` function
2. Centralizing several functions which are used by both Hilbert and Morton distance.  

We could look into this further alongside these follow up issues and for now, simply pass the coords to the `_part1by1` function as type `unit32`? - coords.astype("uint32")

dask_geopandas/core.py
```diff
@@ -336,6 +337,38 @@ def hilbert_distance(self, p=15):
...
+        Parameters
+        ----------
+
+        p : int
+            precision of the Morton curve
+
+        Returns
+        ----------
+        type : pandas.Series
+            Series containing distances from Morton curve
```
martinfleis, line 356: It returns dask Series, not pandas Series.

tastatham: updated in a08e85e

dask_geopandas/core.py
```diff
@@ -343,6 +344,38 @@ def hilbert_distance(self, p=15):
 
         return distances
 
+    def morton_distance(self, p=15):
+
+        """
+        Calculate distance of geoms along Morton curve
```
martinfleis, line 350:
```suggestion
        Calculate the distance of geometries along the Morton curve
        
        The Morton distance can be used to spatially partition Dask-GeoPandas objects,
        by mapping two-dimensional geometries along the Morton space-filing curve.
        
        Each geometry is represented by the midpoint of its bounds and linked to the 
        Morton curve. The function returns a distance from the beginning 
        of the curve to the linked point.
        
        Morton distance is more performant than ``hilbert_distance`` but can result in 
        less optimal partitioning.
```

It is better to have all info in the main docstring.

dask_geopandas/core.py
```diff
@@ -343,6 +344,38 @@ def hilbert_distance(self, p=15):
...
+        Parameters
+        ----------
+
+        p : int
+            precision of the Morton curve
+
+        Returns
+        ----------
+        type : dask.Series
+            Series containing distances from Morton curve
```
martinfleis, line 363:
```suggestion
            Series containing distances along the Morton curve
```

dask_geopandas/core.py
tastatham: thank you

dask_geopandas/morton_distance.py
tastatham: I think this is the last component that needs updating prior to merging it. Any thoughts?

jorisvandenbossche: I would maybe revert the removal of this line for now, and then list it as a follow-up to check / test more.

Otherwise, I would at least eg add a test with a higher p value as mentioned above.

jorisvandenbossche:
The linting is failing (probably something white space in the suggested edit)

dask_geopandas/core.py
```diff
@@ -343,6 +344,48 @@ def hilbert_distance(self, p=15):
...
+        by mapping two-dimensional geometries along the Morton space-filing curve.
+        
+        Each geometry is represented by the midpoint of its bounds and linked to the 
+        Morton curve. The function returns a distance from the beginning 
+        of the curve to the linked point.
+        
+        Morton distance is more performant than ``hilbert_distance`` but can result in 
+        less optimal partitioning.
+
+        The Morton curve is also known as Z-order https://en.wikipedia.org/wiki/Z-order
```
jorisvandenbossche, line 362:
```suggestion
        The Morton curve is also known as Z-order https://en.wikipedia.org/wiki/Z-order.

        The Morton distance can be used to spatially partition Dask-GeoPandas objects,
        by mapping two-dimensional geometries along the Morton space-filing curve.
        
        Each geometry is represented by the midpoint of its bounds and linked to the 
        Morton curve. The function returns a distance from the beginning 
        of the curve to the linked point.
        
        Morton distance is more performant than ``hilbert_distance`` but can result in 
        less optimal partitioning.
```

(I think it's good to "drop" the alternative name in the beginning for visibility)

martinfleis (approved):
I think we're done here. Thanks a lot @tastatham!

brendan-ward (approved):
Thanks for working on this!

Eventually it would be good to add more guidance or at least an appropriate range for the `p` parameter, at least so that the user doesn't overflow data types by using too large a `p` parameter.

dask_geopandas/morton_distance.py
```diff
@@ -0,0 +1,76 @@
+import pandas as pd
+from dask_geopandas.hilbert_distance import _continuous_to_discrete_coords
+
+
+def _morton_distance(gdf, total_bounds, p):
+    """
+    Calculate distance of geoms along Morton curve
```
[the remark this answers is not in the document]
tastatham: thanks Brendan, yes I agree. Hopefully I can find some time to do this before Monday but if not, I will definitely follow up on this.

jorisvandenbossche:
I added back one line in the algorithm, so it matches what pymorton does (see the discussion above https://github.com/geopandas/dask-geopandas/pull/90#discussion_r684612717). We can then properly figure out if it is needed / if we need to cast the integer coordinates to a certain bitsize etc in a follow-up.

jorisvandenbossche:
Thanks @tastatham !

PR #93, opened 2021-08-08
ENH: Implement Geohash

tastatham:
Implement Geohash for Dask-GeoPandas. This provides an alternative (and third) method for spatially partitioning Dask-DataFrames.

Details:
* Calculating the geohash will allow us to partition Dask-GeoPandas using this information.
* The output is returned as a Pandas Series

Currently this pull request uses a vectorized version of the encoder from the `pygeohash` package. I have created a faster encoder based on numpy but this needs debugging first.


dask_geopandas/geohash.py
```diff
@@ -0,0 +1,57 @@
...
+    Returns
+    ---------
+    Pandas Series containing geohash
+    """
+
+    # Calculate bounds
+    bounds = gdf.bounds.to_numpy()
+    # Calculate mid points based on bounds
+    x_mids, y_mids = _calculate_mid_points(bounds)
+    # Vectorize geohash for fast speed up
```
DahnJ, line 27: Just a note, since this comment may suggest otherwise: `np.vectorize` does not actually provide any performance improvements over looping (as per [documentation](https://numpy.org/doc/stable/reference/generated/numpy.vectorize.html)).

dask_geopandas/core.py
```diff
@@ -336,6 +337,32 @@ def hilbert_distance(self, p=15):
 
         return distances
 
+    def geohash(self, precision=15):
+
+        """
+        A function that encodes geometry mid points using Geohash for a given precision
```
martinfleis, line 343: Can you define in the docstring what a "mid point" is?

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,57 @@
+import numpy as np
+import pandas as pd
+from pygeohash import encode
```
martinfleis, line 3: I would rather not have `pygeohash` as a hard dependency. Can you make it optional and raise if it is not available?

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,57 @@
...
+    x_mids, y_mids = _calculate_mid_points(bounds)
+    # Vectorize geohash for fast speed up
+    geohash_vec = np.vectorize(encode)
+    # Encode mid points of geometries using geohash
+    geohash = geohash_vec(y_mids, x_mids, precision)
+
+    return pd.Series(geohash, index=gdf.index, name="geohash")
+
+
+def _calculate_mid_points(bounds):
```
martinfleis, line 35: This is used elsewhere as well. Can you centralise these util functions and reuse the same one?

dask_geopandas/geohash.py
martinfleis: That is correct. @tastatham I'd say keep the implementation but change the comment.

tests/test_geohash.py
```diff
@@ -0,0 +1,66 @@
...
+@pytest.fixture()
+def geoseries_polygons():
+    t1 = Polygon([(0, 3.5), (7, 2.4), (1, 0.1)])
+    t2 = Polygon([(0, 0), (1, 1), (0, 1)])
+    sq1 = Polygon([(0, 0), (1, 0), (1, 1), (0, 1)])
+    sq2 = Polygon([(0, 0), (1, 0), (1, 2), (0, 2)])
+    return geopandas.GeoSeries([t1, t2, sq1, sq2])
+
+
+def geohash_dask(geoseries):
```
martinfleis, line 38: This doesn't really work as a test since it uses the same code.

martinfleis (commented):
> Currently this pull request uses a vectorized version of the encoder from the pygeohash package. I have created a faster encoder based on numpy but this needs debugging first.

Are you planning to use your numpy version here before we merge? Or do you want to keep it as a follow-up? Just to know what to expect, I am fine with both (assuming `pygeohash` will be optional dependency).

dask_geopandas/geohash.py
tastatham: updated in b8f6e38

dask_geopandas/core.py
tastatham: updated in b8f6e38

dask_geopandas/geohash.py
tastatham: Updated in 5aca5f50. I have updated core.py (to not raise import errors) and geohash.py (to raise import errors if called). I think this works ok but I haven't dealt with optional dependencies before so there may be a better way how to handle this?

dask_geopandas/geohash.py
tastatham: yes, I was thinking this function could also be centralized alongside the other functions that have been previously discussed. Something like spatial_partition.py? 

Would it be wise to create a separate PR request to centralize several of these functions?

tests/test_geohash.py
tastatham: yes, I need to update this. I was planning on using pygeohash as the reference for the numpy version but I think the numpy version will become another one of the follow-ups.

dask_geopandas/geohash.py
martinfleis: > yes, I was thinking this function could also be centralized alongside the other functions that have been previously discussed. Something like spatial_partition.py?

Yeah, something like that. maybe with `utils` in that so we know what it is for.

> Would it be wise to create a separate PR request to centralize several of these functions? 

Yeah, once this is merged we can move it around.

martinfleis:
@tastatham what is the issue with your numpy implementation? We may help if you can share the code and describe the problem.

tastatham:
> @tastatham what is the issue with your numpy implementation? We may help if you can share the code and describe the problem.

This is available on my [GSOC Github repo](https://github.com/tastatham/gsoc_dask_geopandas_2021/blob/master/functions/geohash/geohash.py). The issue is that in some cases, the Geohash calculated does not match pygeohash. I have yet to go through this to determine the problem. I think at this stage, it's wise to leave this as a follow-up issue.

tastatham:
> > @tastatham what is the issue with your numpy implementation? We may help if you can share the code and describe the problem.
> 
> This is available on my [GSOC Github repo](https://github.com/tastatham/gsoc_dask_geopandas_2021/blob/master/functions/geohash/geohash.py). The issue is that in some cases, the Geohash calculated does not match pygeohash. I have yet to go through this to determine the problem. I think at this stage, it's wise to leave this as a follow-up issue.

Ok, I have committed the numpy version. The tests run ok but in some cases (see example below) the numpy version and pygeohash do not match.

This works;

```python
import geopandas
import dask_geopandas
import numpy as np
from numpy.testing import assert_array_equal

gdf = geopandas.read_file( geopandas.datasets.get_path('naturalearth_cities') )
ddf = dask_geopandas.from_geopandas(gdf, 1)
precision=12

def pygeohash_expected(gdf, precision):
    from pygeohash import encode
    precision = 12

    bounds = gdf.bounds.to_numpy()
    x_mids, y_mids = dask_geopandas.geohash._calculate_mid_points(bounds)

    geohash_vec = np.vectorize(encode)
    expected = geohash_vec(y_mids, x_mids, precision)
    return expected

result = ddf.geohash(precision).compute()
expected = pygeohash_expected(gdf, precision)

print(result[0:10])
print(expected[0:10])

assert_array_equal(np.array(result), expected)
```

This does not;

```python
import geopandas
import dask_geopandas
import numpy as np
from numpy.testing import assert_array_equal

gdf = geopandas.read_file( geopandas.datasets.get_path('naturalearth_lowres') )
ddf = dask_geopandas.from_geopandas(gdf, 1)
precision=12

def pygeohash_expected(gdf, precision):
    from pygeohash import encode
    precision = 12

    bounds = gdf.bounds.to_numpy()
    x_mids, y_mids = dask_geopandas.geohash._calculate_mid_points(bounds)

    geohash_vec = np.vectorize(encode)
    expected = geohash_vec(y_mids, x_mids, precision)
    return expected

result = ddf.geohash(precision).compute()
expected = pygeohash_expected(gdf, precision)

print(result[0:10])
print(expected[0:10])

assert_array_equal(np.array(result), expected)
```

It is the mid-point of Fiji which is the problem. I'm not sure why, so I need to look into this further.

tests/test_geohash.py
tastatham: since I have now committed the numpy version, this now works ok.

martinfleis:
The issue comes from the `bounds` of the Fiji MultiPolygon which crosses the antimeridian. And according to https://www.movable-type.co.uk/scripts/geohash.html, our result is actually correct for the computed midpoint (which doesn't make much sense in this case to be fair) while the result of `pygeohash` is different. The `7uzyrvpupvrf` hash coming from pygeohash is just next to ours `khbn2j0h0j24` (on a level of precision 7).

Since the movable-type tool agrees with us, I don't think that it is something to fix on our side. Can you check with some other implementation?

tastatham:
> The issue comes from the `bounds` of the Fiji MultiPolygon which crosses the antimeridian. And according to https://www.movable-type.co.uk/scripts/geohash.html, our result is actually correct for the computed midpoint (which doesn't make much sense in this case to be fair) while the result of `pygeohash` is different. The `7uzyrvpupvrf` hash coming from pygeohash is just next to ours `khbn2j0h0j24` (on a level of precision 7).
> 
> Since the movable-type tool agrees with us, I don't think that it is something to fix on our side. Can you check with some other implementation?

Thanks for looking into this, I thought it would be something like this.

The vinsci geohash version produces the same results as the pygeohash package

```python
import geopandas
import dask_geopandas
import numpy as np
from numpy.testing import assert_array_equal

gdf = geopandas.read_file( geopandas.datasets.get_path('naturalearth_lowres') )
#gdf = geopandas.read_file( geopandas.datasets.get_path('naturalearth_cities') )
ddf = dask_geopandas.from_geopandas(gdf, 1)
precision=12

def pygeohash_expected(gdf, precision):
    #!pip install git+https://github.com/KimaruThagna/geohash
    from Geohash import encode
    precision = 12

    bounds = gdf.bounds.to_numpy()
    x_mids, y_mids = dask_geopandas.geohash._calculate_mid_points(bounds)

    geohash_vec = np.vectorize(encode)
    # Encode mid points of geometries using geohash
    expected = geohash_vec(y_mids, x_mids, precision)
    return expected

result = ddf.geohash(precision).compute()
expected = pygeohash_expected(gdf, precision)

print(result[0:10])
print(expected[0:10])

assert_array_equal(np.array(result), expected)
```

martinfleis:
Hmm, then it is 2:2 :D. Is there some canonical geohash implementation (in JavaScript or something?). We can talk about it tonight.

dask_geopandas/core.py
brendan-ward: @tastatham it looks like this got updated for the internal `_geohash()` but not the top level `geohash()` here.

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,221 @@
...
+    ---------
+    type : pandas.Series
+        Series containing geohash
+    """
+
+    # Calculate bounds
+    bounds = gdf.bounds.to_numpy()
+    # Calculate mid points based on bounds
+    x_mids, y_mids = _calculate_mid_points(bounds)
+    # Return coordinates
```
brendan-ward, line 30:
```suggestion
    # Create pairs of x and y midpoints
```

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def encode_geohash(coords, precision):
+    """
+    Calculate geohash based on coordinates for a
+    given precision
+
+    Parameters
+    ----------
+    coords : array_like
```
brendan-ward, line 45:
```suggestion
    coords : array_like of shape (n, 2)
        array of [x, y] pairs
```

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,221 @@
...
+    gs_uint8_mat = _encode_base32(g_uint64)
+    geohash = _encode_unicode(gs_uint8_mat, precision)
+
+    return geohash
+
+
+def _encode_quantize_points(coords):
+    """
+    Quantize coordinates by mapping
+    unit intervals [0, 1] & multiplying by 2^32.
```
brendan-ward, line 65:
```suggestion
    unit intervals [0, 1] and multiplying by 2^32.
```

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,221 @@
...
+    g_uint64 = _encode_into_uint64(quantized_coords)
+    gs_uint8_mat = _encode_base32(g_uint64)
+    geohash = _encode_unicode(gs_uint8_mat, precision)
+
+    return geohash
+
+
+def _encode_quantize_points(coords):
+    """
+    Quantize coordinates by mapping
```
brendan-ward, line 64:
```suggestion
    Quantize coordinates by mapping onto
```

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,221 @@
...
+    unit intervals [0, 1] & multiplying by 2^32.
+
+    Parameters
+    ----------
+    coords : array_like
```
brendan-ward, line 69:
```suggestion
    coords : array_like of shape (n, 2)
        array of [x, y] pairs
```

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,221 @@
...
+        coordinate pairs
+
+    Returns
+    ---------
+    quantized_coords : array_like
```
brendan-ward, line 74:
```suggestion
    array_like of shape (n, 2)
```

I don't think you need to name the output values.

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,221 @@
...
+
+    quantized_coords = _encode_quantize_points(coords)
+    g_uint64 = _encode_into_uint64(quantized_coords)
+    gs_uint8_mat = _encode_base32(g_uint64)
+    geohash = _encode_unicode(gs_uint8_mat, precision)
+
+    return geohash
+
+
+def _encode_quantize_points(coords):
```
brendan-ward, line 62:
```suggestion
def _quantize_points(coords):
```

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,221 @@
...
+def _encode_into_uint64(quantized_coords):
+    """
+    Encode quantized coordinates into uint64
+
+    Implementation based on "Geohash in Golang Assembly"
+    blog (https://mmcloughlin.com/posts/geohash-assembly)
+
+    Parameters
+    ----------
+    quantized_mat : array_like
```
brendan-ward, line 96:
```suggestion
    quantized_coords : array_like of shape (n, 2)
        array of quantized coordinate pairs
```

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,221 @@
...
+        coordinate pairs
+
+    Returns
+    ---------
+    quantized_coords : array_like
```
brendan-ward, line 101:
```suggestion
    array_like of shape (n, 2)
        coordinate pairs encoded to uint64 values
```

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,221 @@
...
+        quantized coordinate pairs
+    """
+
+    __s1 = np.array([(1, 0), (0, 2)], dtype="uint64")
+
+    g_uint64 = np.uint64(quantized_coords)
+    g_uint64 = g_uint64.reshape(-1, 2)
+    g_uint64 = np.bitwise_and(
```
brendan-ward, line 109: I think you could use the same bitwise operators that you used in the Morton curve distance implementation

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,221 @@
...
+    c2 = np.uint8(np.bitwise_and(np.right_shift([g_uint64], 45), mask)).flatten()
+    c1 = np.uint8(np.bitwise_and(np.right_shift([g_uint64], 50), mask)).flatten()
+    c0 = np.uint8(np.bitwise_and(np.right_shift([g_uint64], 55), mask)).flatten()
+
+    return np.column_stack((c0, c1, c2, c3, c4, c5, c6, c7, c8, c9, c10, c11))
+
+
+def _encode_unicode(gs_uint8_mat, precision):
+    """
+    Encode quantized coordinates into uint64
```
brendan-ward, line 168:
```suggestion
    Encode uint64 encoded values into unicode values
```

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,221 @@
...
+
+def _encode_base32(g_uint64):
+    """
+    Encode quantized coordinates into uint64
+
+    Implementation based on "Geohash in Golang Assembly"
+    blog (https://mmcloughlin.com/posts/geohash-assembly)
+    Parameters
+    ----------
+    quantized_mat : array_like
```
brendan-ward, line 139: Here and elsewhere, please make sure that your parameters in the docstring match the parameter names declared above for the function.  Also please consider more human readable parameter names, I'm finding `g_uint64` a bit inscrutable here.

dask_geopandas/utils.py
```diff
@@ -0,0 +1,21 @@
+def _calculate_mid_points(bounds):
+    """
+    Calculate geohash based on the mid points of each geometry
+
+    Parameters
+    ----------
+    gdf : GeoDataFrame
```
brendan-ward, line 7: This docstring does not match the implementation

brendan-ward (changes requested):
Thanks for working on this @tastatham . 

Overall it looks like you have the core implementation in place, but there are a few docstrings that need to be updated to match the implementation or help clarify some of the parameters.

I think you can use the vectorized bitwise operators here like you did with Morton.

I concur with the comments on the call about optionally converting to unicode, and not needing to do so for the values used for partitioning.  Perhaps this could be expressed through 2 top-level functions, `geohash` => unicode values, and `geohash_index` (or something!) => `uint64` values.  Or it could be exposed as a parameter, though I'm less certain about a good name, perhaps `raw=True` => `uint64` and `raw=False` => unicode?, or `format="str"` vs ???

dask_geopandas/geohash.py
tastatham: yes I was thinking this would make things easier to read.

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,224 @@
...
+
+    Returns
+    ---------
+    quantized_coords : array_like
+        quantized coordinate pairs
+    """
+
+    _q = np.array([(2.0 ** 32 / 180, 0), (0, 2.0 ** 32 / (180 * 2))], dtype="float64")
+
+    quantized_coords = coords + np.array((90, 180))
```
jorisvandenbossche, line 82:
```suggestion
    quantized_coords = coords + np.array([90, 180])
```

(small nit, but typically we use square brackets (so list instead of tuple) for manually creating an array. Not that it gives a different result, but just somewhat a convention)

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,224 @@
...
+    Returns
+    ---------
+    array_like of shape (n, 2)
+        coordinate pairs encoded to uint64 values
+        quantized coordinate pairs
+    """
+
+    __s1 = np.array([(1, 0), (0, 2)], dtype="uint64")
+
+    g_uint64 = np.uint64(quantized_coords)
```
jorisvandenbossche, line 110:
```suggestion
    g_uint64 = quantized_coords.astype(np.uint64)
```

(also a bit more typical usage, and thus easier to read for most)

jorisvandenbossche:
> I concur with the comments on the call about optionally converting to unicode, and not needing to do so for the values used for partitioning. Perhaps this could be expressed through 2 top-level functions, `geohash` => unicode values, and `geohash_index` (or something!) => `uint64` values. Or it could be exposed as a parameter, though I'm less certain about a good name, perhaps `raw=True` => `uint64` and `raw=False` => unicode?, or `format="str"` vs ???

I think a keyword would be good (eg the `raw=True/False` as you mention).

Now, to add to this, we still need to convert the integer matrix to some value that can be compared. Because the conversion to unicode is actually two steps: converting the integer matrix of shape (n x 12) into bytes ("S12) array of shape (n, ), and then converting the bytes to string. 
And it seems it's actually the bytes -> strings conversion that is expensive (the ints array -> bytes value is a different view on the same data). So maybe we can return bytes or strings, depending on the keyword?

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,224 @@
+import numpy as np
```
jorisvandenbossche, line 1:
```suggestion
"""
Geohash implementation

The code is originally based on the neathgeohash package, Copyright (c) 2020 Marek Dwulit, MIT License
(https://pypi.org/project/neathgeohash/#description).
That implementation is in turn based on "Geohash in Golang Assembly" blog (https://mmcloughlin.com/posts/geohash-assembly).

"""
import numpy as np
```

I would add something like this at the top of the file (copying code from an MIT licensed package is fine, but so referencing its copyright is the expected thing to do, I think (https://choosealicense.com/licenses/mit/)), and then you can remove the reference to the package and blog post elsewhere (now it's mentioned in multiple docstrings)

dask_geopandas/utils.py
tastatham: Updated in d284934241cf02e493a69c138fad2a65b43d7983

dask_geopandas/geohash.py
tastatham: updated in fca27ecb0590e625f15d7db255b9dcb11312fd74

dask_geopandas/geohash.py
tastatham: updated in fca27ecb0590e625f15d7db255b9dcb11312fd74

dask_geopandas/core.py
```diff
@@ -343,6 +349,32 @@ def hilbert_distance(self, p=15):
 
         return distances
 
+    def geohash(self, precision=15):
```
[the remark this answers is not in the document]
tastatham: ah yes, that's exactly what happened. updated in d7ecc6639b996ec4b187322c43eed3148c9cbf45

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,224 @@
...
+    gs_uint8 = np.where(gs_uint8 == 24, 115, gs_uint8)  # s
+    gs_uint8 = np.where(gs_uint8 == 25, 116, gs_uint8)  # t
+    gs_uint8 = np.where(gs_uint8 == 26, 117, gs_uint8)  # u
+    gs_uint8 = np.where(gs_uint8 == 27, 118, gs_uint8)  # v
+    gs_uint8 = np.where(gs_uint8 == 28, 119, gs_uint8)  # w
+    gs_uint8 = np.where(gs_uint8 == 29, 120, gs_uint8)  # x
+    gs_uint8 = np.where(gs_uint8 == 30, 121, gs_uint8)  # y
+    gs_uint8 = np.where(gs_uint8 == 31, 122, gs_uint8)  # z
+
+    gs_uint8.dtype = np.dtype("|S12")
```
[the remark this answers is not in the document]
tastatham: updated in df72e775c33f4ec98331048b3715ae8dcbbcfdea

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,224 @@
...
+
+    return gs_uint8.flatten().astype("U%s" % precision)
```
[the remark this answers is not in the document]
tastatham: updated in df72e775c33f4ec98331048b3715ae8dcbbcfdea

dask_geopandas/core.py
tastatham: updated in df72e775c33f4ec98331048b3715ae8dcbbcfdea

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,224 @@
...
+    Returns
+    ---------
+    array_like containing geohash
+    """
+
+    # Replacement values
+    gs_uint8 = np.where(gs_uint8_mat == 0, 48, gs_uint8_mat)  # 0
+    gs_uint8 = np.where(gs_uint8 == 1, 49, gs_uint8)  # 1
+    gs_uint8 = np.where(gs_uint8 == 2, 50, gs_uint8)  # 2
+    gs_uint8 = np.where(gs_uint8 == 3, 51, gs_uint8)  # 3
```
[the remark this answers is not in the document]
tastatham: thanks @jorisvandenbossche for the suggestion on the replacement values, that's much cleaner. Updated in 72703a4a68f302119ac721c838b504ef58bbfa2d

dask_geopandas/geohash.py
tastatham: updated in 72703a4a68f302119ac721c838b504ef58bbfa2d

dask_geopandas/geohash.py
tastatham: updated in 72703a4a68f302119ac721c838b504ef58bbfa2d

tastatham:
> > I concur with the comments on the call about optionally converting to unicode, and not needing to do so for the values used for partitioning. Perhaps this could be expressed through 2 top-level functions, `geohash` => unicode values, and `geohash_index` (or something!) => `uint64` values. Or it could be exposed as a parameter, though I'm less certain about a good name, perhaps `raw=True` => `uint64` and `raw=False` => unicode?, or `format="str"` vs ???
> 
> I think a keyword would be good (eg the `raw=True/False` as you mention).
> 
> Now, to add to this, we still need to convert the integer matrix to some value that can be compared. Because the conversion to unicode is actually two steps: converting the integer matrix of shape (n x 12) into bytes ("S12) array of shape (n, ), and then converting the bytes to string.
> And it seems it's actually the bytes -> strings conversion that is expensive (the ints array -> bytes value is a different view on the same data). So maybe we can return bytes or strings, depending on the keyword? (I don't know if there is an easy way to express the 12 uint8 integer values as a single integer number, while viewing them as bytes basically solves that)

yes I agree, I'll look into this tomorrow with a fresh head.

tastatham:
> > > I concur with the comments on the call about optionally converting to unicode, and not needing to do so for the values used for partitioning. Perhaps this could be expressed through 2 top-level functions, `geohash` => unicode values, and `geohash_index` (or something!) => `uint64` values. Or it could be exposed as a parameter, though I'm less certain about a good name, perhaps `raw=True` => `uint64` and `raw=False` => unicode?, or `format="str"` vs ???
> > 
> > 
> > I think a keyword would be good (eg the `raw=True/False` as you mention).
> > Now, to add to this, we still need to convert the integer matrix to some value that can be compared. Because the conversion to unicode is actually two steps: converting the integer matrix of shape (n x 12) into bytes ("S12) array of shape (n, ), and then converting the bytes to string.
> > And it seems it's actually the bytes -> strings conversion that is expensive (the ints array -> bytes value is a different view on the same data). So maybe we can return bytes or strings, depending on the keyword? (I don't know if there is an easy way to express the 12 uint8 integer values as a single integer number, while viewing them as bytes basically solves that)
> 
> yes I agree, I'll look into this tomorrow with a fresh head.

I've modified geohash.py, core.py, and test_geohash.py to take the raw argument in 44cb91bab2b573575a1c1b695a071c5a1c8c0ba5

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,245 @@
...
+        dtype="uint8",
+    )
+
+    encoded_base32 = replacement[encoded_base32]
+
+    if raw is True:
+        encoded_base32 = encoded_base32.view(np.dtype("|S12"))
+        return encoded_base32.flatten().astype(f"U{precision}")
+
+    elif raw is not True:
```
jorisvandenbossche, line 244:
```suggestion
    if raw:
        encoded_base32 = encoded_base32.view(np.dtype("|S12"))
        return encoded_base32.flatten().astype(f"U{precision}")
    else:
```

(although `raw is True` is certainly more explicit, it's general practice to exclude the `is True`. See eg https://www.python.org/dev/peps/pep-0008/#programming-recommendations, one of the bullet points "Don't compare boolean values to True or False using ==")

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,245 @@
...
+            121,
+            122,
+        ],
+        dtype="uint8",
+    )
+
+    encoded_base32 = replacement[encoded_base32]
+
+    if raw is True:
+        encoded_base32 = encoded_base32.view(np.dtype("|S12"))
```
jorisvandenbossche, line 241: This line should probably be done before the if/else? I think we should always returns a single 1D array, and only have the final conversion to unicode depend on the `raw` keyword.

jorisvandenbossche (commented):
Thanks for the updates! Added some more comments

setup.py
```diff
@@ -5,13 +5,13 @@
 
 install_requires = [
     "geopandas",
+    "numpy<1.21,>=1.17",
```
martinfleis, line 8: Why `<1.21`? If possible, we should not restring the maximum version, only the minimum.

dask_geopandas/geohash.py
tastatham: Updated in 9947e80

setup.py
tastatham: The latest version of numba (v.0.54.0) specifies [numpy<1.21,>=1.17](https://pypi.org/project/numba/). Alternatively, we could use an earlier version of numba? v.0.53.1 specifies [NumPy >=1.15 (can build with 1.11 for ABI compatibility)](https://pypi.org/project/numba/0.53.1/)

dask_geopandas/core.py
```diff
@@ -389,6 +395,40 @@ def morton_distance(self, p=15):
...
+        Parameters
+        ----------
+        precision : int
+            precision of the Geohash
+
+        Returns
+        ----------
+        type : pandas.Series
+            Series containing Geohash
+        """
```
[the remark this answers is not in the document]
tastatham: updated in 7c2afc2

dask_geopandas/core.py
```diff
@@ -17,6 +17,12 @@
 from .morton_distance import _morton_distance
 
 
+try:
+    from .geohash import _geohash
+except ImportError:
+    external_module = None
```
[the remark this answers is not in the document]
tastatham: updated in 3adf523

dask_geopandas/core.py
```diff
@@ -389,6 +395,40 @@ def morton_distance(self, p=15):
...
+            raise ValueError(
+                "The Geohash precision only accepts an integer value between 1 and 12"
+            )
+
+        # Calculate hilbert distances for each partition
+        geohashes = self.map_partitions(
+            _geohash,
+            precision=precision,
+            raw=raw,
+            meta=pd.Series([], name="geohash", dtype="str"),
```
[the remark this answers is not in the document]
tastatham: gotcha, updated in 7dbd0e5

setup.py
martinfleis: If numba needs certain versions let's leave it to them to take care of pins. We just list `numba`.

dask_geopandas/core.py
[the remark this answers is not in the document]
martinfleis: @tastatham just a minor note - you don't need to comment with a link to a commit. Just hit Resolve conversation once you're done with the suggestion.

setup.py
tastatham: `numba` was throwing a dependency error when trying to install Dask-GeoPandas.

martinfleis:
(ignore the readthedocs failure in checks, I turned it on for #110)

dask_geopandas/geohash.py
tastatham: thanks, I need to familiarise myself better with PEP 8

tastatham:
I have updated geohash in f10e539e971de5442a74d6980b62ba8abc9d31a5
If we decompose the `encode_geohash` function, there are four individual components;
1. `encode_quantize_points` - array_like of shape (n, 2)
2. `_encode_into_uint64` - array_like of shape (n, 2)
3. `_encode_base32` - array_like of shape (n, 12)
4. `_encode_unicode` - array_like of shape (n, precision)

For the sake of partitioning, we can stop after encoding the quantized points into uint64 (steps 1 and 2) through spreading and bit-interleaving the integers - returning the `integer geohash`. However, if we also want to return the `string geohash` (for later use), we must also run steps 3 and 4.

dask_geopandas/core.py
```diff
@@ -389,6 +390,41 @@ def morton_distance(self, p=15):
...
 
+    def geohash(self, string=False, p=12):
+
+        """
+        Calculate geohash based on the middle points of the geometry bounds
+        for a given precision
+
+        Parameters
+        ----------
+        string : bool
```
martinfleis, line 401:
```suggestion
        string : bool (default False)
```

dask_geopandas/core.py
```diff
@@ -389,6 +390,41 @@ def morton_distance(self, p=15):
...
+            to return string or int Geohash
+        p : int
```
martinfleis, line 403:
```suggestion
        p : int (default 12)
```

dask_geopandas/core.py
```diff
@@ -389,6 +390,41 @@ def morton_distance(self, p=15):
 
         return distances
 
+    def geohash(self, string=False, p=12):
```
martinfleis, line 393: Shall we default to `False` here? I guess that most users will expect standard string-based geohash out of this method and we can set it to `False` easily manually in `spatial_shuffle`.

martinfleis (commented):
A few more comments in code. I think we're almost there.

One thing I am not sure about is the `utils.py` module. We should certainly make it private if we want it (i.e. as `_utils.py`) but it may be actually fine to keep these shared methods in one of the existing files and just import in other from there.

dask_geopandas/core.py
tastatham: that makes sense.

tastatham:
> A few more comments in code. I think we're almost there.
> 
> One thing I am not sure about is the `utils.py` module. We should certainly make it private if we want it (i.e. as `_utils.py`) but it may be actually fine to keep these shared methods in one of the existing files and just import in other from there.

The idea behind the utils.py file was to create a module that included several functions which were shared across the different partitioning methods. In hindsight, this should have been part of a small & separate PR.

I have temporarily renamed utils.py to _utils.py. We can discuss whether we want to keep _utils.py on the call later on.

dask_geopandas/core.py
```diff
@@ -389,6 +390,38 @@ def morton_distance(self, p=15):
...
 
+    def geohash(self, string=True, p=12):
+
+        """
+        Calculate geohash based on the middle points of the geometry bounds
+        for a given precision
+
+        Parameters
+        ----------
+        string : bool (default True)
```
brendan-ward, line 401:
```suggestion
        as_string : bool (default True)
```

dask_geopandas/core.py
```diff
@@ -389,6 +390,38 @@ def morton_distance(self, p=15):
...
+        if p not in range(1, 13):
+                raise ValueError(
+                    "The Geohash precision only accepts an integer value between 1 and 12"
+                )
+
+        geohashes = self.map_partitions(
+            _geohash,
+            string=string,
+            p=p,
+            meta=pd.Series([], name="geohash", dtype=object),
```
brendan-ward, line 420: Is the `dtype` here dependent on whether or not string or integer values are being returned?

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,248 @@
...
+    Returns
+    ---------
+    geohash: array containing either int or string
+        geohashes for each mid point
+    """
+
+    quantized_coords = _encode_quantize_points(coords)
+    int_geohash = _encode_into_uint64(quantized_coords)
+
+    if string is True:
```
brendan-ward, line 72: Suggestion: the conditional can be simplified a bit since it involves return statements:

```
if not string:
    return int_geohash

gs_uint8_mat = _encode_base32(int_geohash)
str_geohash = _encode_unicode(gs_uint8_mat, p)
return str_geohash
```

dask_geopandas/core.py
```diff
@@ -389,6 +390,38 @@ def morton_distance(self, p=15):
 
         return distances
 
+    def geohash(self, string=True, p=12):
+
+        """
+        Calculate geohash based on the middle points of the geometry bounds
+        for a given precision
+
```
brendan-ward, line 398: Add something like

```
Only geographic coordinates (longitude, latitude) are supported.
```

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,248 @@
...
+    # Interleave x and y bits so that x and y occupy even and odd bit levels
+    x = x[:, 0] | x[:, 1]
+    x = x >> 4
+
+    return x
+
+
+def _encode_base32(encoded_uint64):
+    """
+    Encode quantized coordinates into base32 pairs
```
brendan-ward, line 149:
```suggestion
    Encode quantized coordinates into base32 pairs.
    
```

martinfleis:
@tastatham can you make sure that CI pass?

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,248 @@
...
+
+    Parameters
+    ----------
+    g_uint64 : array_like
+        coordinate pairs encoded to uint64 values
+
+    Returns
+    ---------
+    array_like of shape (n, 12)
+        with base 32 values of type unasigned integer
```
brendan-ward, line 164:
```suggestion
        with base 32 values as 8-bit unsigned integers
```

brendan-ward (commented):
A few more minor edits to consider

dask_geopandas/core.py
tastatham: good spot! updated

dask_geopandas/geohash.py
```diff
@@ -0,0 +1,245 @@
...
+The vectorized implementation for quantization and bit interleaving is in turn based on,
+"Geohash in Golang Assembly" blog (https://mmcloughlin.com/posts/geohash-assembly).
+
+"""
+import numpy as np
+import pandas as pd
+from .utils import _calculate_mid_points
+
+
+# Implementation based on deprecated https://pypi.org/project/neathgeohash/#description
```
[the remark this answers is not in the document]
jorisvandenbossche: Can you do this one

brendan-ward (approved):
Thanks @tastatham for all your work here.  I gave this a quick review and didn't see anything major left to resolve.

martinfleis (approved):
Thanks!

Issue #96, opened 2021-08-18
read_parquet does not retain `crs`


TomAugspurger:
Currently, read_parquet loses the `crs` of the geometry column(s).

```python
In [1]: import geopandas, dask_geopandas

In [2]: df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))

In [3]: df.to_parquet("test.parquet")

In [4]: dask_geopandas.read_parquet("test.parquet").geometry.crs  # None
```

I'm looking into how best to handle this. I think all the information is there in the parquet file.

jorisvandenbossche:
Whoops, I would have thought that the `assert_geodataframe_equal` would have checked for the CRS, and that this was covred. But so it does do that, but the issue is that we are testing the computed result. And at that point, the resulting geopandas GeoDataFrame actually has the correct CRS .. (because at that points it gets it from the actual geometry array stored in the geometry column), but its the dask-geopandas GeoDataFrame that doesn't have the crs set (because here it's getting it from the `meta`, as you fixed in the PR). 

Thanks for the catch!

PR #97, opened 2021-08-18
BUG: read CRS in read_parquet

TomAugspurger:
This sets the CRS when reading back a file that was written with the `geo-arrow-spec`'s metadata.

It's moves the conversion of the Dask `meta` dataframe down to a private method unfortunately, but AFAICT, that's the most convenient place to get access to the Parquet Dataset, so that we can read the `.schema.metadata`. I didn't look too closely.

Closes https://github.com/geopandas/dask-geopandas/issues/96

### DIFF
--- a/dask_geopandas/io/parquet.py
+++ b/dask_geopandas/io/parquet.py
@@ -58,11 +58,6 @@ class GeoArrowEngine(ArrowEngine):
     def read_metadata(cls, *args, **kwargs):
         meta, stats, parts, index = super().read_metadata(*args, **kwargs)
 
-        # Update meta to be a GeoDataFrame
-        # TODO convert columns based on GEO metadata (this will now only work
-        # for a default "geometry" column)
-        meta = geopandas.GeoDataFrame(meta)
-
         # get spatial partitions if available
         regions = geopandas.GeoSeries([_get_partition_bounds(part) for part in parts])
         if regions.notna().all():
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Whoops, I have pre-commit setup now and fixed the linting issue.

jorisvandenbossche:
> It's moves the conversion of the Dask `meta` dataframe down to a private method unfortunately, but AFAICT, that's the most convenient place to get access to the Parquet Dataset, so that we can read the `.schema.metadata`. I didn't look too closely.

We could bring this up with dask to more "formatlize" that method to make subclassing the engine easier (a bit similar as there are the `_arrow_table_to_pandas`/`_pandas_to_arrow_table` methods). 

I think the current solution in your PR is fine for now. But I am wondering if we can longer-term combine it with the parsing of the bbox from the metadata (to populate `spatial_partitions`):

https://github.com/geopandas/dask-geopandas/blob/99fa1a865156649d66c9ae4fa6fe119b6be6206c/dask_geopandas/io/parquet.py#L67

This is not actually an ideal implementation, as it incurs file IO for each of the partitions to get the Parquet metadata (and possibly doing that again, as dask might also already have done this, if it gathered statistics without a `_metadata` file).

TomAugspurger:
> This is not actually an ideal implementation, as it incurs file IO for each of the partitions to get the Parquet metadata (and possibly doing that again, as dask might also already have done this, if it gathered statistics without a _metadata file).

I was just looking into that section :) I believe that it's currently not working with some (all?) remote filesystems like Azure Blob Storage:

```pytb
In [1]: import dask_geopandas

In [2]: dask_geopandas.read_parquet("abfs://gbif/occurrence/2021-08-01/occurrence.parquet", storage_options={'account_name': 'ai4edataeuwest'})
...
FileNotFoundError: [Errno 2] Failed to open local file 'gbif/occurrence/2021-08-01/occurrence.parquet/000000'. Detail: [errno 2] No such file or directory
```

We try to reopen just the fragment path there, but don't provide the filesystem / open it with the filesystem.

I'm hopeful https://github.com/dask/dask/issues/8058 will help with the issue of having to (re)-read the per-file metadata. That GBIF file has >1,000 partitions :)

jorisvandenbossche:
> We try to reopen just the fragment path there, but don't provide the filesystem / open it with the filesystem.

Ah, thanks for that catch. Yes, we definitely need to use the filesystem there. Will look into that separately.

> I'm hopeful dask/dask#8058 will help with the issue of having to (re)-read the per-file metadata. T

I am not fully sure, but need to take a good look at the discussion there and materialize my thoughts ;)

jorisvandenbossche:
Let's merge this for now, since it's fixing the bug. Thanks Tom!

jorisvandenbossche:
> Yes, we definitely need to use the filesystem there. Will look into that separately.

Started with adding s2/moto tests that replicate the error -> https://github.com/geopandas/dask-geopandas/pull/103

PR #104, opened 2021-08-22
ENH: Add shuffle

tastatham:
Implement shuffle for Dask-GeoPandas. This allows for spatially shuffling/partitioning a Dask-GeoPandas object using one of the three methods (hilbert, morton or geohash) or a user defined column.


dask_geopandas/core.py
```diff
@@ -389,6 +389,60 @@ def morton_distance(self, p=15):
 
         return distances
 
+    def shuffle(self, by="hilbert", column=None, drop=True, npartitions=10, **kwargs):
```
martinfleis, line 392: It needs to be called differently (`spatial_shuffle`?), dask already has the `shuffle` method.

dask_geopandas/core.py
```diff
@@ -389,6 +389,60 @@ def morton_distance(self, p=15):
...
+        """
+        A function that spatially shuffles a Dask-GeoSeries object by a method
+        or a user-defined column
+
+        Parameters
+        ----------
+        by : str
+            partitioning method
+        column : str
+            precomputed column
```
martinfleis, line 403: I don't think we need this. Using precomputed column is essentially `set_index`.

dask_geopandas/core.py
```diff
@@ -389,6 +389,60 @@ def morton_distance(self, p=15):
...
+        drop : bool
+            to drop the partitioning information held in index
+        npartitions : int
+            number to partition DataFrame
+
```
martinfleis, line 408: `**kwargs` are missing from the docstring.

Also, I would probably expose `p` directly and pass `**kwargs` to `set_index`.

dask_geopandas/core.py
```diff
@@ -389,6 +389,60 @@ def morton_distance(self, p=15):
...
+            ddf = self.set_index(by, drop=drop, sorted=False, npartitions=npartitions)
+
+        elif column is not None:
+            ddf = self.set_index(
+                column, drop=drop, sorted=True, npartitions=npartitions
+            )
+
+        # Calculate convex hull of each partition
+        ddf.calculate_spatial_partitions()
```
martinfleis, line 436: Can we control this via keyword defaulting to `True`? I can imagine cases when I don't want to do that.

dask_geopandas/core.py
```diff
@@ -389,6 +389,60 @@ def morton_distance(self, p=15):
...
+
+        # Optional drop index
+        if drop is True:
+            ddf = ddf.reset_index()
+        else:
+            pass
```
martinfleis, line 442:
```suggestion
```

No need for `else` clause if it is empty.

martinfleis (commented):
Thanks for looking into that.

It may also be useful to explore `shuffle` method of dask.dataframe (https://github.com/dask/dask/blob/4229c16cf0cc7ed5f60f313618680bfd04e381ea/dask/dataframe/shuffle.py#L301). We may be able to use that instead of `set_index` to avoid some overhead.

martinfleis:
Just dropping here a note from @mrocklin from https://github.com/dask/dask/issues/8075#issuecomment-903803077 for a reference

> Also, as a heads-up, I'm working on a newer algorithm here.  The API will
still be the same (I saw that you're doing this for geopandas) but there
may be new algorithms to learn from in the future that may perform slightly
better at large scale.

gjoseph92:
I also recommend @TomAugspurger's approach. Calling into `super().shuffle(...)` seems better than fully overriding the method. Plus, there are some improvements to shuffling coming up that it would be easier to benefit from with this method.

Seems like if `on` is None, then you want `super().shuffle(self, hash_method(self), ...)`, otherwise just `super().shuffle(self, on, ...)`. (Note that shuffling also isn't quite the same as `set_index`.)

martinfleis:
@tastatham To get closer to the actual release, I have opened #131 to implement `spatial_shuffle` based on `set_index` there. That should free a bit of your time and you can finish that notebook with documentation to wrap up GSoC.

martinfleis:
Superseded by #131

PR #107, opened 2021-08-27
BUG: fix read_parquet with a non-default geometry column name

### LINKED ISSUE #105: BUG: parquet with renamed geometry is not read as GeoDataFrame
Having a parquet saved via geopandas with a non-default geometry column name breaks dask-geopandas IO as it falls back to DataFrame (but still properly converts WKB to geometry).

```py
import geopandas
import dask_geopandas

df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
df = df.rename_geometry("geom")
df.to_parquet("df.pq")
dask_geopandas.read_parquet("tst/df.pq")

Dask DataFrame Structure:
              pop_est continent    name  iso_a3 gdp_md_est    geom
npartitions=1                                                     
                int64    object  object  object    float64  object
                  ...       ...     ...     ...        ...     ...
Dask Name: read-parquet, 1 tasks
```

jorisvandenbossche:
Closes #105

### DIFF
--- a/dask_geopandas/io/parquet.py
+++ b/dask_geopandas/io/parquet.py
@@ -73,14 +73,18 @@ class GeoArrowEngine(ArrowEngine):
         )
         if schema.metadata and b"geo" in schema.metadata:
             geo_meta = json.loads(schema.metadata[b"geo"])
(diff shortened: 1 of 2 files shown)


martinfleis (approved):
Thanks!

PR #108, opened 2021-08-27
BUG: fix read_parquet for multiple geometry columns

jorisvandenbossche:
Can be done as follow-up on https://github.com/geopandas/dask-geopandas/pull/107

~TODO: This doesn't yet restore the correct CRS for a different column~ This is also done now


martinfleis (approved):
Looks good, thanks!

PR #109, opened 2021-09-09
Documentation: Update Z-order wikipedia link page reference

jtbaker:
Cool project. While reading through some of the source code (and learning about some of the concepts) I noticed that the wikipedia page in the documentation was for a graphics concept and not the mathematics one. Perhaps the wiki page originally routed to a disambiguation page when the doc-string was first created.


martinfleis:
Thanks!

PR #111, opened 2021-09-10
BUG: correctly set CRS of spatial_partitions in read_parquet

jorisvandenbossche:
Encountered this while trying the sjoin benchmarks from spatialpandas, getting warnings about mismatching CRS in sjoin.


martinfleis (approved):
Thanks!

PR #118, opened 2021-09-27
BUG: fix set_geometry with a dask_geopandas.GeoSeries

### LINKED ISSUE #117: set_geometry error with unhasable type Point
When I try to set the geometry on a points geodataframe as shown in the readme I get a `TypeError: unhashable type: Point`.

I am using:
 - pandas 1.3.3
 - geopandas 0.9.0
 - dask-geopandas 0.1.0 (installed from master branch last week)
 - dask 2021.7.2
 - pygeos 0.10.2

# Example:
```Python
import numpy as np
import pandas as pd
import dask.dataframe as dd
import dask_geopandas

n = 1000
X = np.random.random((n, 2))
X[:, 0] = 360 * X[:, 0] - 180
X[:, 1] = 170 * X[:, 1] - 85
df = pd.DataFrame(X, columns=['lon', 'lat'])

ddf = dd.from_pandas(df, npartitions=2)
ddf = ddf.set_geometry(dask_geopandas.points_from_xy(ddf, 'lon', 'lat'))
```

Just as a side note: I tried to work around it using:
```Python
from pygeos.lib import points

ddf["geometry"] = points(ddf[["lon", "lat"]].values)
ddf = ddf.set_geometry("geometry")
```

Which works if I do not use a `dask.distributed` cluster otherwise I get a deserialization issue (I use `pygeos.lib.points` here because the conveniance function to create points in pygeos executes eagerly). This seems to be a special issue with points though because the second example works for `pygeos.box` geometries.

jorisvandenbossche:
Closes #117

### DIFF
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -419,7 +419,10 @@ class GeoDataFrame(_Frame, dd.core.DataFrame):
         # calculate ourselves to use meta and not meta_nonempty, which would
         # raise an error if meta is an invalid GeoDataFrame (e.g. geometry
         # column name not yet set correctly)
-        meta = self._meta.set_geometry(col)
+        if isinstance(col, GeoSeries):
+            meta = self._meta.set_geometry(col._meta)
+        else:
+            meta = self._meta.set_geometry(col)
         return self.map_partitions(M.set_geometry, col, meta=meta)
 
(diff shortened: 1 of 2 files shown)


martinfleis (approved):
Thanks for the quick fix!

PR #119, opened 2021-10-29
Update Parquet implementation for compatibility with latest dask

jorisvandenbossche:
This fixes the Parquet tests with the last dask release.

It was already said in https://github.com/geopandas/dask-geopandas/pull/97 that we unfortunately needed to dig into a private dask API, and so that has now changed. We should still see to bring it up with dask if this can be made "semi-public" (for developers), but this patch should already fix it (and it's not that much code to keep compatibility with both older and newest dask).


martinfleis (approved):
Thanks!

@jorisvandenbossche can you open an issue in dask to discuss this? You know the best what is needed from their side.

PR #121, opened 2021-11-03
BUG: do not copy spatial partitions when None

### LINKED ISSUE #120: _Frame tries to copy spatial_partitions even when it is None.
```python 
import geopandas as gpd
import dask_geopandas as dgpd
from shapely.geometry import Point

gdf = gpd.GeoSeries([Point(0, 0), Point(1, 1)])
dgdf = dgpd.from_geopandas(gdf, npartitions=2)
dgdf.copy()

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-57-2e1e1df5b72e> in <module>
      5 gdf = gpd.GeoSeries([Point(0, 0), Point(1, 1)])
      6 dgdf = dgpd.from_geopandas(gdf, npartitions=2)
----> 7 dgdf.copy()

/opt/anaconda/envs/apollo/lib/python3.8/site-packages/dask_geopandas/core.py in copy(self)
    169         """
    170         self_copy = super().copy()
--> 171         self_copy.spatial_partitions = self.spatial_partitions.copy()
    172         return self_copy
    173 

AttributeError: 'NoneType' object has no attribute 'copy'
```

This should be an easy fix, i'll make a PR for it if everyone is happy with just checking if it's None before copying it.

bernardpazio:
check if spatial partitions is none before copying to fix #120

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -447,6 +447,11 @@ def test_map_partitions_get_geometry(geodf_points):
     assert_geoseries_equal(result, expected)
 
 
+def test_copy_none_spatial_partitions(geoseries_points):
+    ddf = dask_geopandas.from_geopandas(geoseries_points, npartitions=2)
+    ddf_copy = ddf.copy()
+
+
 class TestDissolve:
     def setup_method(self):
         self.world = geopandas.read_file(
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -449,7 +449,9 @@ def test_map_partitions_get_geometry(geodf_points):
 
 def test_copy_none_spatial_partitions(geoseries_points):
     ddf = dask_geopandas.from_geopandas(geoseries_points, npartitions=2)
+    ddf.spatial_partitions = None
     ddf_copy = ddf.copy()
+    assert ddf_copy.spatial_partitions is None
 
 
 class TestDissolve:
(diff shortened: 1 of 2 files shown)


tests/test_core.py
```diff
@@ -447,6 +447,11 @@ def get_geometry(partition):
     assert_geoseries_equal(result, expected)
 
 
+def test_copy_none_spatial_partitions(geoseries_points):
+    ddf = dask_geopandas.from_geopandas(geoseries_points, npartitions=2)
+    ddf_copy = ddf.copy()
+
```
jorisvandenbossche, line 453:
```suggestion
    assert ddf.spatial_partitions is None
    ddf_copy = ddf.copy()
    assert ddf_copy.spatial_partitions is None

```

(not strictly needed, but making it a bit more explicit)

jorisvandenbossche (commented):
Thanks for the PR! Added a small comment, otherwise looks good

tests/test_core.py
bernardpazio: yeah that is fair, I've gone with explicitly setting it to None, just in case something happens in between which makes it not None at some point.

martinfleis (approved):
Thanks!

(let's wait for #119 to make CI happy before merging)

jorisvandenbossche:
Thanks again @bernardpazio !

PR #125, opened 2021-11-12
CI: remove the pyproj 2.6 pin

jorisvandenbossche:
In the meantime there are pyproj 3.1+ releases that have fixed the issue in pyproj 3.0 (https://github.com/geopandas/dask-geopandas/pull/36#issuecomment-791898795), so we can remove the global 2.6 pin. I noted in https://github.com/geopandas/dask-geopandas/pull/123 however that this seems to give a windows failure related to S3 / AWS (I suppose triggered by the packages updates, by no longer pinning to an old pyproj version and thus also old versions of other packages).

### DIFF
--- a/continuous_integration/envs/37-latest.yaml
+++ b/continuous_integration/envs/37-latest.yaml
@@ -9,7 +9,7 @@ dependencies:
   - distributed
   - geopandas
   - pygeos
-  - pyproj=2.6  # unpin when pyproj 3.1 is released with multithreading fix
+  - pyproj=2.6
   # test dependencies
   - pytest
   - pytest-cov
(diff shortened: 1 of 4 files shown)


martinfleis:
Do you have any idea where does the error come from? It looks that all our tests pass and then it raises this error before it starts the next action.

jorisvandenbossche:
> It looks that all our tests pass and then it raises this error before it starts the next action.

I suppose the crash happens during "clean-up" of the python process (so after pytest printed that everything is passing)

> What also puzzles me is that it is an AWS error, I thought that GHA runs on Azure.
Merge state

I suppose that is not related. We have aws-c-io installed in the environment, I am just not fully sure which package might be using that (we have S3 tests, but I thought s3fs / boto are not relying on the AWS C++ packages).

jorisvandenbossche:
It was a conda-forge build issue, which should be fixed now: https://github.com/conda-forge/arrow-cpp-feedstock/issues/567

I clean-up my temporary commits, let's see what CI says now.

jorisvandenbossche:
Passing now!

PR #126, opened 2021-11-18
Pass through `crs` when using `points_from_xy`

gadomski:
This enables the usage [in the readme](https://github.com/geopandas/dask-geopandas/blob/master/README.rst#example) to have a CRS on the resultant data frame (which is needed for writing to parquet).

Sidecar fixes:
- Swaps 'longitude' and 'latitude' in the readme as they were incorrect
- Adds a `requirements-dev.txt` file that contains the packages needed to run the test suite locally

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -60,7 +60,7 @@ data.
 
    ddf = dask_geopandas.from_dask_dataframe(ddf)
    ddf = ddf.set_geometry(
-       dask_geopandas.points_from_xy(ddf, 'latitude', 'longitude')
+       dask_geopandas.points_from_xy(ddf, 'longitude', 'latitude')
    )
 
 Writing files (and reading back) is currently supported for the Parquet file
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -135,6 +135,21 @@ def test_points_from_xy():
     assert_geodataframe_equal(result, expected)
 
 
+def test_points_from_xy_with_crs():
+    latitude = [40.2, 66.3]
+    longitude = [-105.1, -38.2]
+    expected = geopandas.GeoSeries(
+        geopandas.points_from_xy(longitude, latitude, crs="EPSG:4326")
+    )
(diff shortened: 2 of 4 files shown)


README.rst
```diff
@@ -60,7 +60,7 @@ data.
 
    ddf = dask_geopandas.from_dask_dataframe(ddf)
    ddf = ddf.set_geometry(
-       dask_geopandas.points_from_xy(ddf, 'latitude', 'longitude')
+       dask_geopandas.points_from_xy(ddf, 'longitude', 'latitude')
```
jorisvandenbossche, line 63: Whoops .. thanks for the catch!

tests/test_core.py
```diff
@@ -135,6 +135,19 @@ def test_points_from_xy():
...
+    latitude = [40.2, 66.3]
+    longitude = [-105.1, -38.2]
+    expected = geopandas.points_from_xy(longitude, latitude, crs="EPSG:4326")
+    df = pd.DataFrame({"longitude": longitude, "latitude": latitude})
+    ddf = dd.from_pandas(df, npartitions=2)
+    actual = dask_geopandas.points_from_xy(
+        ddf, "longitude", "latitude", crs="EPSG:4326"
+    )
+    assert isinstance(actual, dask_geopandas.GeoSeries)
+    assert list(actual) == list(expected)
```
jorisvandenbossche, line 148: This won't in itself verify the that the crs is set, so maybe you can change this with `assert_geoseries_equal(actual.compute(), expected)`

jorisvandenbossche (commented):
Thanks for the PR! One small comment on the test

tests/test_core.py
gadomski: Updated in 88f03cf.

jorisvandenbossche (approved):

martinfleis (approved):
Thanks!

PR #128, opened 2021-12-30
DEV: rename master to main

jorisvandenbossche:
xref https://github.com/geopandas/community/issues/7

This is the only occurrence of "master" that I could find.


martinfleis (approved):
Thanks!

Issue #59, opened 2021-06-17
BUG: set_index results in invalid dask GeoDataFrame (partitions are DataFrames)


jorisvandenbossche:
Using the dask `set_index` method results in an "invalid" dask_geopandas.GeoDataFrame, where the partitions are no longer GeoDataFrames but DataFrames (which then results in errors when computing spatial operations)

```python
import geopandas
import dask_geopandas

gdf = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
ddf = dask_geopandas.from_geopandas(gdf, npartitions=4)

>>> type(ddf.partitions[0].compute())
geopandas.geodataframe.GeoDataFrame

>>> ddf2 = ddf.set_index("continent")
# still a dask_geopandas GeoDataFrame
>>> type(ddf2)   
dask_geopandas.core.GeoDataFrame
# but partitions under the hood no longer a geopandas.GeoDataFrame
>>> type(ddf2.partitions[0].compute())   
pandas.core.frame.DataFrame
# which results in errors when doing actual computations
>>> ddf2.area.compute()
....
AttributeError: 'DataFrame' object has no attribute 'area'
```

DahnJ:
Interestingly, this is not the case for all the partitions:
```python
>>> [type(partition.compute()) for partition in ddf2.partitions]

[pandas.core.frame.DataFrame,
 geopandas.geodataframe.GeoDataFrame,
 pandas.core.frame.DataFrame,
 pandas.core.frame.DataFrame]
```

jorisvandenbossche:
Ah, interesting. Then it might also be a bug in GeoPandas (if some operation on a GeoDataFrame results in a pandas DataFrame where it could have preserved the GeoDataFrame type)

DahnJ:
I believe the problem is in the fact that Dask uses [`partd.pandas`](https://github.com/dask/partd/blob/main/partd/pandas.py) to (de)serialize the dataframe:

```python
from partd.pandas import serialize, deserialize
df = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))
type(deserialize(serialize(df)))  # pandas.core.frame.DataFrame
```

[...]

PR #131, opened 2022-01-17
ENH: implement `spatial_shuffle` method

### LINKED ISSUE #88: ENH: Implement `spatial_shuffle`
We should have a top-level method on `dask_geopandas.GeoDataFrame` and `dask_geopandas.GeoSeries` allowing to shuffle data into spatially coherent partitions. (xref #8)

Let's assume we have `hilbert_distance` and `geohash` functions based on which we can spatially repartition gdf. I would then like to have a single method (`spatial_shuffle`?) that can consume both.

```py
# using hilbert
ddf = ddf.spatial_shuffle('hilbert')

# using geohash
ddf = ddf.spatial_shuffle('geohash')
```

This API can be extended by other methods without the necessity to have two methods for each algorithm (esp. since the `repartition_` methods will be largely the same).

_Originally posted by @martinfleis in https://github.com/geopandas/dask-geopandas/issues/71#issuecomment-876260579_

_note: moving to a separate specific issue to keep track of GSoC._

martinfleis:
This supersedes #104 and closes #88.

~I still need to add tests.~

I am not allowing custom Series in `by` as in that case you would just use `set_index` directly.


dask_geopandas/core.py
```diff
@@ -429,6 +432,101 @@ def geohash(self, string=True, p=12):
 
         return geohashes
 
+    def spatial_shuffle(
+        self,
+        by="hilbert",
+        p=None,
+        calculate_partitions=True,
+        drop_index=True,
```
tastatham, line 440: [drop_index](https://docs.dask.org/en/latest/_modules/dask/dataframe/core.html#DataFrame.set_index) is not necessary because we are not using a column to partition the Dask GeoDataFrame - so there is nothing to drop.

martinfleis: This controls whether the original index is dropped or not. If you have some meaningful info there, you may want to keep it. If it's just a range, you can drop it. See how it is passed below to `set_index`.

dask_geopandas/core.py
```diff
@@ -429,6 +432,100 @@ def geohash(self, string=True, p=12):
...
+        calculate_partitions=True,
+        npartitions=None,
+        divisions=None,
+        **kwargs,
+    ):
+        """
+        Shuffle the data into spatially consistent partitions.
+
+        This realigns the dataset to be spatially sorted, i.e. geometries that are
+        spatially near each other together will be within the same partition. This is
```
jorisvandenbossche, line 448:
```suggestion
        spatially near each other will be within the same partition. This is
```

dask_geopandas/core.py
```diff
@@ -429,6 +432,100 @@ def geohash(self, string=True, p=12):
...
+        The spatial information is stored in the index and will replace the existing
+        index.
+
+        Note that ``spatial_shuffle`` uses ``set_index`` under the hood and comes with
+        all its potential performance drawbacks.
+
+        Parameters
+        ----------
+        by : string (default 'hilbert')
+            sorting method, one of {'hilbert', 'morton, 'geohash'}. See
```
jorisvandenbossche, line 461:
```suggestion
            Spatial sorting method, one of {'hilbert', 'morton', 'geohash'}. See
```

dask_geopandas/core.py
```diff
@@ -429,6 +432,100 @@ def geohash(self, string=True, p=12):
...
+            ``hilbert_distance``, ``morton_distance`` and ``geohash`` methods for
+            details.
+        p : int (default None)
+            precision of the sorting method. Defaults to 15 for Hilbert and Morton
```
jorisvandenbossche, line 465:
```suggestion
            Precision of the sorting method. Defaults to 15 for Hilbert and Morton
```

dask_geopandas/core.py
```diff
@@ -429,6 +432,100 @@ def geohash(self, string=True, p=12):
...
+            curves and 12 for geohash.
+        calculate_partitions : bool (default True)
+            calculate new spatial partitions after shuffling
```
jorisvandenbossche, line 468:
```suggestion
            Calculate new spatial partitions after shuffling
```

dask_geopandas/core.py
```diff
@@ -429,6 +432,100 @@ def geohash(self, string=True, p=12):
...
+            The ideal number of output partitions. If None, use the same as the input.
+            If 'auto' then decide by memory use. Only used when divisions is not given.
+            If divisions is given, the number of output partitions will be
+            len(divisions) - 1.
+        divisions: list, optional
+            The “dividing lines” used to split the new index into partitions. Needs to
+            match the values returned by the sorting method.
+
+        **kwargs
+            keyword arguments passed to ``set_index``.
```
jorisvandenbossche, line 479:
```suggestion
            Keyword arguments passed to ``set_index``.
```

dask_geopandas/core.py
```diff
@@ -429,6 +432,100 @@ def geohash(self, string=True, p=12):
...
+        else:
+            raise ValueError(
+                f"'{by}' is not supported. Use one of ['hilbert', 'morton, 'geohash']."
+            )
+
+        if "shuffle" in kwargs:
+            shuffle = kwargs.pop("shuffle")
+            if shuffle != "tasks":
+                warnings.warn(
+                    "Temporarily forcing the 'tasks' shuffle mode due to an upstream issue."
```
jorisvandenbossche, line 511: Do we have an issue about this?

dask_geopandas/core.py
```diff
@@ -429,6 +432,100 @@ def geohash(self, string=True, p=12):
...
+        Note that ``spatial_shuffle`` uses ``set_index`` under the hood and comes with
+        all its potential performance drawbacks.
+
+        Parameters
+        ----------
+        by : string (default 'hilbert')
+            sorting method, one of {'hilbert', 'morton, 'geohash'}. See
+            ``hilbert_distance``, ``morton_distance`` and ``geohash`` methods for
+            details.
+        p : int (default None)
```
jorisvandenbossche, line 464: Not for this PR, but I am wondering if we shouldn't rename this for all those methods to `precision` for clarity

dask_geopandas/core.py
jorisvandenbossche: Ah, this is https://github.com/geopandas/dask-geopandas/issues/59

dask_geopandas/core.py
```diff
@@ -429,6 +432,100 @@ def geohash(self, string=True, p=12):
...
+
+        Returns
+        -------
+        dask_geopandas.GeoDataFrame
+
+        Notes
+        -----
+        This method, similarly to ``calculate_spatial_partitions``, is computed
+        partially eagerly as it needs to load the active geometry column and compute
+        spatial partitions.
```
jorisvandenbossche, line 489: This is the case when `calculate_spatial_partitions=True`? If so, maybe mention in the above paragraph you can make it fully lazily by passing that keyword.  

And for example, I suppose that if you do a directly following write to disk, such as `ddf.spatial_shuffle().to_parquet()`, you actually want to turn this off? Because then dask can combine all things in a single graph, and we will calculate the bounds of each partition anyway when writing to Parquet.

jorisvandenbossche (commented):
Looking good!

dask_geopandas/core.py
jorisvandenbossche: We might still mention that in eg the notes section that this will remove any existing index? 

Do we still want to option to not _set_ the distance as the index? Or you can do a manual `.reset_index(drop=True)` in case you don't want it? 
Eg when writing the result of `spatial_shuffle` to a partitioned (parquet) dataset, you might not necessarily care about the exact distance values for later analyses, and then it just occupies extra space in the files.

martinfleis: > We might still mention eg in the notes section that this will remove any existing index?

It is on the line 452.

> Do we still want to option to not set the distance as the index? Or you can do a manual `.reset_index(drop=True)` in case you don't want it?

I would vote for the latter. If we added the keyword, we would in the end only used it to call `.reset_index(drop=True)` under the hood.

dask_geopandas/core.py
martinfleis: That is a good point.

dask_geopandas/core.py
martinfleis: > This is the case when calculate_spatial_partitions=True? If so, maybe mention in the above paragraph you can make it fully lazily by passing that keyword.

I don't think so, it seems to be computed eagerly no matter the `calculate_spatial_partitions` option because dask needs to know the divisions for the new index. And to get those, it needs to load geometry and compute Hilbert curve.

jorisvandenbossche: Yeah, the `set_index` documentation (https://docs.dask.org/en/latest/generated/dask.dataframe.DataFrame.set_index.html) has this about the `compute` keyword:

> Whether or not to trigger an immediate computation. Defaults to False. Note, that even if you set `compute=False`, an immediate computation will still be triggered if `divisions` is `None`.

jorisvandenbossche: Maybe we can then clarify the note a little bit? Eg "... as it needs to calculate the distances for all existing partitions before it can determine the divisions for the new spatially-shuffled partitions"

martinfleis:
I still have to debug the CI failures but they seem to be related to the geohash implementation, not `spatial_shuffle`.

martinfleis:
I have fixed the dtype issue in `geohash` affecting Windows envs and added tests for all sorting methods using the real-world data. 

The final failure caused by this PR is in `37-latest` and is causes by incorrect sorting of dask dataframe, that was fixed in dask 2021.03.0. So I'd skip that for now.

martinfleis:
CI is as green as it gets right now. Consider this PR ready.

jorisvandenbossche:
> remove p from goehash (unused)

That's because it is only used when being converted to a string value?

martinfleis:
> That's because it is only used when being converted to a string value?

Yes, which is not used here.

jorisvandenbossche (commented):
Looks good!

I merged the set_index fix, so you should be able to clean that up here before merging

jorisvandenbossche (approved):

jorisvandenbossche:
> move to GeoDataFrame only 

It doesn't work for GeoSeries out of the box? (it's certainly less useful / less typical to do it with a GeoSeries though)

martinfleis:
> It doesn't work for GeoSeries out of the box? (it's certainly less useful / less typical to do it with a GeoSeries though)

GeoSeries / Series does not have `set_index` in Dask apparently.

jorisvandenbossche:
Thanks!

PR #134, opened 2022-01-23
REF: vectorized numpy-based Hilbert Curve implementation

jorisvandenbossche:
See https://github.com/geopandas/geopandas/pull/2297 (the original motivation was to have a numpy-based implementation for geopandas (to avoid adding a dependency on numba), but since it turns out to work well, we can use it here just as well)

This is a numpy-based implementation of the distance along the Hilbert Curve, giving the same results as the numba-based version we already have, but thus not using numba but a vectorized implementation based on the version in GEOS (which is basically https://github.com/rawrunprotected/hilbert_curves)

With the version of GEOS / https://github.com/rawrunprotected/hilbert_curves, it was actually trivially to have it run vectorized on a numpy array of discrete coordinates as well (literally the same code, with only a few small edits to convert it into python code).

Based on a first test, this new implementation gives the same result as the current numba-based one, while actually being faster (so in addition of not requiring numba as a dependency). On 1 million points, I get ca 500ms for the numba-based implementation vs ~80ms for this numpy-based implementation. 
(this is only for the lower-level encoding of discrete coordinates to distance, so the actual difference starting from a GeoSeries will be smaller) 
Notebook: https://notebooksharing.space/view/fb83908941339cc7654dbe271c02b2df49eaf8cc4d434077da58066f9cd1ff86#displayOptions=

cc @martinfleis @tastatham


martinfleis (approved):
I still have no idea how that bit shifting magic works but this PR looks good!

Thanks!

jorisvandenbossche:
I kept the keyword rename for a separate PR (https://github.com/geopandas/dask-geopandas/pull/144), since that also affects the other distance methods

martinfleis (approved):
Thanks!

Issue #135, opened 2022-01-23
Failing tests with geopandas and dask main


jorisvandenbossche:
There are two failing tests at the moment in the dev build (eg https://github.com/geopandas/dask-geopandas/runs/4912124074?check_suite_focus=true).

In an environment with just dask main, one of the two tests fails: `tests/test_core.py::TestDissolve::test_split_out`. This is caused by https://github.com/dask/dask/pull/8468 which was merged yesterday.

In an environment with just geopandas main, the other test fails: `tests/test_core.py::test_set_geometry_with_dask_geoseries` with

```
______________________________________________________________________________________ test_set_geometry_with_dask_geoseries ______________________________________________________________________________________

    def test_set_geometry_with_dask_geoseries():
        df = pd.DataFrame({"x": [0, 1, 2, 3], "y": [1, 2, 3, 4]})
        dask_obj = dd.from_pandas(df, npartitions=2)
        dask_obj = dask_geopandas.from_dask_dataframe(dask_obj)
>       dask_obj = dask_obj.set_geometry(dask_geopandas.points_from_xy(dask_obj, "x", "y"))

tests/test_core.py:356: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Dask DataFrame Structure:
                   x      y
npartitions=2              
0              int64  int64
2                ...    ...
3                ...    ...
Dask Name: GeoDataFrame, 4 tasks
key = 'set_geometry'

    def __getattr__(self, key):
        if key in self.columns:
            return self[key]
        else:
>           raise AttributeError("'DataFrame' object has no attribute %r" % key)
E           AttributeError: 'DataFrame' object has no attribute 'set_geometry'

../dask/dask/dataframe/core.py:4268: AttributeError
```

Given the error that might be related to the changes in geopandas related to `_constructor` returning DataFrame vs GeoDataFrame depending geometry columns being present or not. Update: indeed starts to fail with https://github.com/geopandas/geopandas/pull/2060

[...]

PR #136, opened 2022-01-23
BUG: fix ddf.set_geometry for geopandas main

jorisvandenbossche:
See https://github.com/geopandas/dask-geopandas/issues/135#issuecomment-1019492684

This adds a `set_geometry` to dask.dataframe, similar as we do that in GeoPandas. However, that patching is also not super clean, so we might want to avoid it?

It is to fix this code snippet:

```
ddf = dd.from_pandas(df, npartitions=2)
gddf = ddf.set_geometry(dask_geopandas.points_from_xy(ddf, "x", "y"))
```

Basically, in the above, the line `dask_geopandas.from_dask_dataframe(ddf)` has become "useless", as it still returns a dd.DataFrame. 

Instead of patching dd.DataFrame with `set_geometry`, another option could be to have the user do something like `dask_geopandas.from_dask_dataframe(ddf, geometry=dask_geopandas.points_from_xy(ddf, "x", "y"))`


martinfleis:
> Instead of patching dd.DataFrame with set_geometry, another option could be to have the user do something like dask_geopandas.from_dask_dataframe(ddf, geometry=dask_geopandas.points_from_xy(ddf, "x", "y"))

This is exactly what I wanted to suggest. With the snippet above, the `ddf` after `dask_geopandas.from_dask_dataframe(ddf)` would be invalid anyway as it does not contain geometry. We should try to eb consistent with geopandas/geopandas#2060.

jorisvandenbossche:
> With the snippet above, the `ddf` after `dask_geopandas.from_dask_dataframe(ddf)` would be invalid anyway as it does not contain geometry

Note that it is not actually an "invalid" dask_geopandas.GeoDataFrame, as it simply returns a "proper" dask.dataframe.DataFrame with pd.DataFrames as partitions. But of course that also makes the function a useless no-op (if there was not yet a geometry dtyped column).

jorisvandenbossche:
You can compare it in this way: in geopandas, you can also create a GeoDataFrame in two ways if you start from a pandas.DataFrame:

```python
gdf = df.set_geometry(geometries)
# vs
gdf = geopandas.GeoDataFrame(df, geometry=geometries)
```

And in `dask_geopandas`, `from_dask_dataframe` is a bit the equivalent of the `geopandas.GeoDataFrame(..)` constructor in this case. 

Previously, you could also do this in geopandas:

```
gdf = geopandas.GeoDataFrame(df)
gdf = gdf.set_geometry(geometries)
```

this still works on geopandas main, but the first line will also actually return a DataFrame (if `df` doesn't have a geometry dtyped column), and the second like actually creates the GeoDataFrame and only works because geopandas patches pd.DataFrame with a `set_geometry` method.

martinfleis:
> Note that it is not actually an "invalid" dask_geopandas.GeoDataFrame, as it simply returns a "proper" dask.dataframe.DataFrame with pd.DataFrames as partitions

Which surely is not what you expect from it. It essentially returns what you passed in.

martinfleis:
I actually didn't know we can use `DataFrame.set_geometry`. It kind of feels wrong but I don't mind that.

We may actually want to do both here. Adding `set_geometry` to dask.DataFrame as you just did and expanding `from_dask_dataframe` to require `geometry` (to make sure it returns `dask_geopandas.GeoDataFrame`.

jorisvandenbossche:
Updated to include both.

martinfleis (approved):
Thanks for the changes!

PR #139, opened 2022-01-27
DOC: documentation follow-ups

martinfleis:
A follow-up on #110 

- inheriting proper docstrings
- adding missing methods to API
- moving notebooks from `notebooks` to user guide
- cleaning ReadMe
- pinning numpydoc to avoid a (already fixed) bug in 1.2.0


README.rst
```diff
@@ -16,6 +15,32 @@ or with financial support from you or your company.
...
+Documentation
+-------------
+
+See the documentation on https://dask-geopandas.readthedocs.io/en/latest/
+
+Installation
+------------
+
+This package depends on GeoPandas and Dask. In addition, it is recommended to
+install PyGEOS, to have faster spatial operations and enable multithreading. See
```
brendan-ward, line 27: suggestion:
`In addition, it is recommended to install PyGEOS, to have faster` => `You can also install PyGEOS to have faster...`

jorisvandenbossche: I would maybe phrase it a bit stronger than "you can also install", as it is really strongly recommended no? (I don't think there is much point in using dask-geopandas without it ..)

doc/source/guide/basic-intro.ipynb
```diff
@@ -17,17 +17,14 @@
     "import geopandas\n",
     "\n",
     "import dask.dataframe as dd\n",
-    "import dask_geopandas\n",
-    "\n",
-    "import pyproj\n",
-    "pyproj.datadir.set_data_dir(\"/home/joris/miniconda3/envs/geo-dev/share/proj/\")\n",
```
jorisvandenbossche, line 23: Good catch! :)

README.rst
martinfleis: Well, this sentence is incorrect anyway. We depend on `pygeos` in `calculate_spatial_partitions`.

PR #140, opened 2022-01-27
add warning about split_out issue in dissolve

martinfleis:
Adding a warning about the issue mentioned in #135 and https://github.com/dask/dask/issues/8611

It is unlikely that we will be able to fix this on the geopandas side.


martinfleis:
Fixed on the Dask side.

PR #144, opened 2022-01-28
Rename level/precision keywords for Hilbert/Morton/GeoHash

jorisvandenbossche:
xref https://github.com/geopandas/geopandas/pull/2297#discussion_r794040784

Started with the public method in core.py, once we agree on the naming / docs, will adjust in the lower level bits as well.

For Hilbert and Morton, it certainly makes sense to use `level`. For Geohash I am a bit less sure, there it feels that `precision` is better suited.

cc @martinfleis @brendan-ward @tastatham


dask_geopandas/core.py
```diff
@@ -318,8 +318,7 @@ def explode(self):
     def cx(self):
         return _CoordinateIndexer(self)
 
-    def hilbert_distance(self, p=15):
-
+    def hilbert_distance(self, level=15):
```
jorisvandenbossche, line 321: I have also been wondering about this default of 15 here, since the max allowed value (and thus the best precision) is 16. So maybe the intention was to use 16?

Now it is an iterative algorithm, I am also wondering if this actually matters that much (another reason to use the highest precision by default)

I think eg PostGIS simply harcodes to use 16 (see https://github.com/postgis/postgis/blob/18ce69feba33b13d9fc6122dd594669d9b1bb977/liblwgeom/lwinline.h#L260, it's based on exactly the same bit-shifting implementation, and when level=16, you can leave it out. For example, in our code we have `x = x << (16 - level)` at the start of this function, and with level=16 this simplifies to `x = x << 0`, which is a no-op)

martinfleis:
> For Hilbert and Morton, it certainly makes sense to use level. For Geohash I am a bit less sure, there it feels that precision is better suited.

Yes, that sound about right. We will just need to find a keyword that covers both these options to be used in `spatial_shuffle`.

dask_geopandas/core.py
martinfleis: @tastatham do you remember that?

tastatham: From memory, there was no significant overhead in selecting the highest precision we set  - which was 15.
I'm not sure why we selected this value, maybe it was based on their SpatialPandas recommendations? I need to look into my notes/notebooks.

As an iterative algorithm, the numba version does allow a precision greater than 16, whereas the maximum p value for the bit-shifting algorithm is currently 16.
The bit shifting could be easily updated to support higher p values (e.g. up to 32/64) but I haven't found a scenario where a value greater than 15/16 is necessary for encoding spatial data. 
So long story short, setting max p=16 is fine.

jorisvandenbossche: Indeed, the max of 16 is determined by the use of uint32. By eg using uint64 we could go with a higher precision if that use case would ever come up. But let's stick with 16 for now then.

brendan-ward:
Agreed on `level` for Hilbert / Morton and `precision` for Geohash.

dask_geopandas/core.py
```diff
@@ -402,15 +403,17 @@ def geohash(self, string=True, p=12):
         ----------
         as_string : bool (default True)
```
jorisvandenbossche, line 404: The docstring says ``as_string``, but the signature has `string=True/False`. Does somebody remember which of the two we prefer?

(`as_string` might be a bit more explicit?)

martinfleis: I like `as_string` more but don't remember the discussion.

jorisvandenbossche:
I am going to merge this, so I can finalize https://github.com/geopandas/dask-geopandas/pull/134

Issue #137, opened 2022-01-24
BUG: parquet IO fails if one or more partitions are empty


martinfleis:
In some cases, like after a spatial join, some of your partitions may end up being empty. If that happens, dask_geopandas raises an RuntimeError.

```py
import geopandas
import dask_geopandas

import warnings

warnings.filterwarnings('ignore', message='.*initial implementation of Parquet.*')

world = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
cities = geopandas.read_file(geopandas.datasets.get_path("naturalearth_cities"))
africa = world.loc[world.continent == "Africa"]

ddf_cities = dask_geopandas.from_geopandas(cities, npartitions=12)

hilbert = ddf_cities.hilbert_distance()
ddf_cities = ddf_cities.set_index(hilbert, shuffle="tasks")

african_cities = dask_geopandas.sjoin(ddf_cities, africa)

african_cities.to_parquet("tst/")
```

```py
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
~/mambaforge/envs/geo_dev/lib/python3.9/site-packages/dask/dataframe/io/parquet/arrow.py in _append_row_groups(metadata, md)
     51     try:
---> 52         metadata.append_row_groups(md)
     53     except RuntimeError as err:

~/mambaforge/envs/geo_dev/lib/python3.9/site-packages/pyarrow/_parquet.pyx in pyarrow._parquet.FileMetaData.append_row_groups()

RuntimeError: AppendRowGroups requires equal schemas.

The above exception was the direct cause of the following exception:

RuntimeError                              Traceback (most recent call last)
/var/folders/2f/fhks6w_d0k556plcv3rfmshw0000gn/T/ipykernel_60072/2656726537.py in <module>
     17 african_cities = dask_geopandas.sjoin(ddf_cities, africa)
     18 
---> 19 african_cities.to_parquet("tst/")

~/Git/dask-geopandas/dask_geopandas/core.py in to_parquet(self, path, *args, **kwargs)
    584             index=data.index,
    585         )
--> 586 
    587     return df.map_partitions(
    588         func, x, y, z, meta=geopandas.GeoSeries(), token="points_from_xy"

[...]

PR #147, opened 2022-01-28
BUG: fix read_parquet with empty partitions (all-NaN bbox)

jorisvandenbossche:
Address the second (reading) part of https://github.com/geopandas/dask-geopandas/issues/137

### DIFF
--- a/tests/io/test_parquet.py
+++ b/tests/io/test_parquet.py
@@ -138,3 +138,21 @@ def test_parquet_roundtrip_s3(s3_resource, s3_storage_options):
     assert result.crs == df.crs
     # reading back also populates the spatial partitioning property
     assert result.spatial_partitions is not None
+
+
+def test_parquet_empty_partitions(tmp_path):
+    df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
+    # Creating filtered dask dataframe with at least one empty partition
+    ddf = dask_geopandas.from_geopandas(df, npartitions=4)
+    ddf_filtered = ddf[ddf["pop_est"] > 1_000_000_000]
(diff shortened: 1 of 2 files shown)


tests/io/test_parquet.py
```diff
@@ -138,3 +138,21 @@ def test_parquet_roundtrip_s3(s3_resource, s3_storage_options):
...
+    assert (ddf_filtered.map_partitions(len).compute() == 0).any()
+
+    basedir = tmp_path / "dataset"
+    # TODO don't write metadata file as that fails with empty partitions on
+    # inferring the schema
+    ddf_filtered.to_parquet(basedir, write_metadata_file=False)
+
+    result = dask_geopandas.read_parquet(basedir)
+    assert_geodataframe_equal(result.compute(), df[df["pop_est"] > 1_000_000_000])
+    # once one partition has no spatial extent, we don't restore the spatial partitions
```
martinfleis, line 157: Why we do this? In the case of empty partitions, we can simply have `None` in the GeoSeries and it still helps a lot. Is it a safe option in case of somehow incorrectly created parquet where non-empty partitions are missing bounds? Shall we maybe offer a way around that for cases like this one?

martinfleis: I understand the consequences, I just think it may be worth doing something a bit smarter here now. But let's leave it for another PR.

jorisvandenbossche: Yes, especially with the current situation that we easily write empty partitions to parquet (which then will result in this kind of bboxes), this is quite annoying. 

(but for this specific case, the better solution might also be to not write empty partitions to parquet)

PR #148, opened 2022-01-30
CI: use mamba on CI

martinfleis:
Use official `conda-incubator` action and mambaforge. The current setup is super slow, this should help it. It mirrors geopandas/geopandas settings.


martinfleis:
Cool, this works now and CI fail much faster than before :).

jorisvandenbossche (approved):
Thanks!

Issue #130, opened 2022-01-13
GeoPandas feature parity


jorisvandenbossche:
Quickly exploring which spatial methods that are defined in GeoPandas are not yet available here:

```python
import pandas as pd
import geopandas
import dask.dataframe as dd
import dask_geopandas

methods_pandas = set([n for n in dir(pd.DataFrame) if not n.startswith("_")])
methods_geopandas = set([n for n in dir(geopandas.GeoDataFrame) if not n.startswith("_")])
methods_dask = set([n for n in dir(dd.DataFrame) if not n.startswith("_")])
methods_dask_geopandas = set([n for n in dir(dask_geopandas.GeoDataFrame) if not n.startswith("_")])

methods_geopandas_extra = methods_geopandas - methods_pandas
methods_dask_geopandas_extra = methods_dask_geopandas - methods_dask

>>> methods_geopandas_extra - methods_dask_geopandas_extra
{'cascaded_union',
 'covered_by',
 'covers',
 'estimate_utm_crs',
 'explore',
 'from_features',
 'from_file',
 'from_postgis',
 'has_sindex',
 'iterfeatures',
 'overlay',
 'rename_geometry',
 'sjoin',
 'sjoin_nearest',
 'to_file',
 'to_postgis',
 'to_wkb',
 'to_wkt'}

>>> methods_dask_geopandas_extra - methods_geopandas_extra
{'calculate_spatial_partitions',
 'hilbert_distance',
 'interpolate',
 'morton_distance',
 'set_geometry',
 'to_dask_dataframe'}
```

Some quick first notes:

- `covers` and `covered_by` are 2 missing predicates that should be trivial to add here
- `sjoin` was added as a method in geopandas, we should do the same here

martinfleis:
One more thing to this is also API parity, e.g. `sjoin` in geopandas now uses `predicate` while here we still have `op` only.

jorisvandenbossche:
An updated version:

```python
>>> methods_geopandas_extra - methods_dask_geopandas_extra
{'cascaded_union',
 'clip_by_rect',
 'estimate_utm_crs',
 'explore',
 'from_features',
 'from_file',
 'from_postgis',
 'has_sindex',
 'iterfeatures',
 'overlay',
 'sjoin_nearest',
 'to_file',
```

I think from this list, `overlay` and `sjoin_nearest` are the most useful (but also complicated). 
[...]

PR #149, opened 2022-01-30
ENH: match geopandas sjoin and clip API

### LINKED ISSUE #127: clip calls wrong clip function
Hey all.

I am loving the work on this project!
I've stumbled upon a bug and wanted you to let you know about it.
I don't know if it is actually worth "fixing".
If one calls `.clip` on a geopandas dask dataframe object, the dask clip function (clipping values between `min/max` is called)

To reproduce:

```python
import geopandas
import dask_geopandas

path_to_data = geopandas.datasets.get_path("nybb")
gdf = geopandas.read_file(path_to_data)
ddf = dask_geopandas.from_geopandas(gdf, npartitions=2)
ddf.clip(gdf.geometry[0])
```

martinfleis:
Closes #127, xref #130

- add `clip` as a GeoDataFrame method
- support GeoSeries in `clip` and add it the method
- add `sjoin` as a GeoDataFrame method
- change `op` to `predicate` in `sjoin`

### DIFF
--- a/dask_geopandas/sjoin.py
+++ b/dask_geopandas/sjoin.py
@@ -1,3 +1,5 @@
+import warnings
+
 import numpy as np
 import geopandas
 
@@ -7,7 +9,7 @@ from dask.highlevelgraph import HighLevelGraph
 from .core import from_geopandas, GeoDataFrame
 
 
-def sjoin(left, right, how="inner", op="intersects"):
+def sjoin(left, right, how="inner", predicate="intersects", **kwargs):
     """
     Spatial join of two GeoDataFrames.
 
@@ -19,7 +21,7 @@ def sjoin(left, right, how="inner", op="intersects"):
         partitioning information).
     how : string, default 'inner'
         The type of join. Currently only 'inner' is supported.
-    op : string, default 'intersects'
+    predicate : string, default 'intersects'
         Binary predicate how to match corresponding rows of the left and right
         GeoDataFrame. Possible values: 'contains', 'contains_properly',
         'covered_by', 'covers', 'crosses', 'intersects', 'overlaps',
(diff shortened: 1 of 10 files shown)


dask_geopandas/sjoin.py
```diff
@@ -38,6 +40,14 @@ def sjoin(left, right, how="inner", op="intersects"):
...
     of the left and right GeoDataFrame.
     """
+    if "op" in kwargs:
+        predicate = kwargs.pop("op")
+        deprecation_message = (
+            "The `op` parameter is deprecated and will be removed"
+            " in a future release. Please use the `predicate` parameter"
+            " instead."
+        )
+        warnings.warn(deprecation_message, FutureWarning)
```
jorisvandenbossche, line 50:
```suggestion
        warnings.warn(deprecation_message, FutureWarning, stacklevel=2)
```

dask_geopandas/sjoin.py
```diff
@@ -46,8 +56,8 @@ def sjoin(left, right, how="inner", op="intersects"):
     if isinstance(right, geopandas.GeoDataFrame):
         right = from_geopandas(right, npartitions=1)
 
-    name = "sjoin-" + tokenize(left, right, how, op)
-    meta = geopandas.sjoin(left._meta, right._meta, how=how, op=op)
+    name = "sjoin-" + tokenize(left, right, how, predicate)
+    meta = geopandas.sjoin(left._meta, right._meta, how=how, predicate=predicate)
```
jorisvandenbossche, line 60: This essentially makes us require geopandas>=0.10 I think? (which is fine for me, but maybe add the lower bounds in setup.py)

jorisvandenbossche (commented):
Thanks!

dask_geopandas/sjoin.py
martinfleis: Ah, true. I'll change the requirements. I think that there's no harm having versions of geopandas and dask-geopandas closely coupled, i.e. we can always require latest geopandas with the latest dask-geopandas.

jorisvandenbossche: Yes, and especially now for releasing the first non-alpha release, it's fine to require the latest geopandas release.

PR #150, opened 2022-01-31
TST: skip dissolve split_out tests for dask 2022.01.1

jorisvandenbossche:
We know that this test will always be failing for this exact version of dask, assuming it gets fixed on the dask side in the next version (https://github.com/dask/dask/issues/8611). So then let's skip it to get current CI a bit easier to interpret.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -4,6 +4,7 @@ import pandas as pd
 import numpy as np
 import geopandas
 from shapely.geometry import Polygon, Point, LineString, MultiPoint
+import dask
 import dask.dataframe as dd
 from dask.dataframe.core import Scalar
 import dask_geopandas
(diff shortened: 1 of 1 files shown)


tests/test_core.py
```diff
@@ -507,12 +508,30 @@ def test_sum(self):
...
         dd_split = self.ddf.dissolve("continent", split_out=4)
         assert dd_split.npartitions == 4
         assert_geodataframe_equal(gpd_default, dd_split.compute(), check_like=True)
 
+    @pytest.mark.skipif(
+        Version(dask.__version__) == Version("2022.01.1"),
+        reason="Regression in dask 2022.01.1 https://github.com/dask/dask/issues/8611",
+    )
+    @pytest.mark.xfail
+    def test_split_out_name(self):
```
jorisvandenbossche, line 526: I added an equivalent test with a different geometry name (since this has come up in the discussion as a case that could fail with certain solutions). Adding this custom one for now, but we might want to see if we can improve the workflow here, eg with some parametrization on the dataset, so this can more easily be done on multiple tests.

jorisvandenbossche:
Going to merge to get CI green

tests/test_core.py
martinfleis: That is a good point. We should test renamed geometries extensively as it will likely be a pain point, given how dask treats dataframes internally.

PR #151, opened 2022-01-31
Add covers and covered_by methods

jorisvandenbossche:
xref https://github.com/geopandas/dask-geopandas/issues/130


martinfleis (approved):
That was easy. Thanks!

martinfleis:
One more thing. Can you add it to the API docs?

PR #153, opened 2022-02-09
CI: use strict channel priority to avoid GDAL from main channel

jorisvandenbossche:
The tests on python 3.7 are currently failing due to a fiona import error (https://github.com/geopandas/dask-geopandas/runs/5123735878?check_suite_focus=true), which I suppose is caused by a mix from default / conda-forge channel (for some reason, it started to install GDAL 3.4 from the main channel instead of GDAL 3.2 from conda-forge, since one of the last days)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -51,6 +51,7 @@ jobs:
           miniforge-version: latest
           miniforge-variant: Mambaforge
           use-mamba: true
+          channel-priority: strict
 
       - name: Check and Log Environment
         run: |


martinfleis:
That didn't help for some reason. It still comes from default.

jorisvandenbossche:
Yeah, currently trying to reproduce it locally

jorisvandenbossche:
I can reproduce it locally, opened https://github.com/mamba-org/mamba/issues/1465

martinfleis:
This seems to have fixed itself, but we have merged #91 with a bit incompatible with our minimal version (I'd bump it as I said elsewhere).

jorisvandenbossche:
> but we have merged https://github.com/geopandas/dask-geopandas/pull/91 with a bit incompatible with our minimal version (I'd bump it as I said elsewhere).

Yeah, I was ignoring the failures on the 37 build .. 

Closing this, as I think for mamba it shouldn't actually be necessary. I assume it was caused by some temporary incorrect metadata issue.

PR #154, opened 2022-02-10
DOC: Add spatial partitioning notebook

tastatham:
I have added a draft spatial partitioning notebook.
This demonstrates the behaviour of `spatial_shuffle` and some best practices.
For now, I haven't included a benchmark between a Dask-GeoDataFrame that has and hasn't been spatially partitioned.
We can agree on whether this would be best placed here or somewhere else.


martinfleis:
Can you move the notebook to `doc/source/guide` now that the #139 is merged? And add a link to `guide.md`?

jorisvandenbossche:
@tastatham thanks for working on this! This is becoming a nice notebook. 

In addition to the comment about the dataset above, some other general comments:

- Regarding the "What precision?" section:
  - The `level` for Hilbert: now it doesn't use an iterative algorithm, the performance is actually not influenced anymore by the level, I think. 
  - When using geohash within `spatial_shuffle`, the level/precision is currently ignored (so specifying a keyword for this doesn't do anthing), because we use the integer return value, and the precision is only used when converting to a string
  - Given that, I am not sure how useful it is to talk about this aspect, and I would maybe simplify this section a bit (can of course still show the three different methods next to each other, but then the section can be about that instead of about the precision)
- The `calculate_partitions=True` and `divisions=None` are the default values in `spatial_shuffle`, so to reduce the line length and improve the readability of the code snippets, I would remove those arguments from each `spatial_shuffle` call

tastatham:
> @tastatham thanks for working on this! This is becoming a nice notebook.
> 
> In addition to the comment about the dataset above, some other general comments:
> 
> * Regarding the "What precision?" section:
>   
>   * The `level` for Hilbert: now it doesn't use an iterative algorithm, the performance is actually not influenced anymore by the level, I think.
>   * When using geohash within `spatial_shuffle`, the level/precision is currently ignored (so specifying a keyword for this doesn't do anthing), because we use the integer return value, and the precision is only used when converting to a string
>   * Given that, I am not sure how useful it is to talk about this aspect, and I would maybe simplify this section a bit (can of course still show the three different methods next to each other, but then the section can be about that instead of about the precision)
> * The `calculate_partitions=True` and `divisions=None` are the default values in `spatial_shuffle`, so to reduce the line length and improve the readability of the code snippets, I would remove those arguments from each `spatial_shuffle` call

Thanks @jorisvandenbossche for reviewing the notebook.

- I have updated the example dataset, which now uses Contiguous United States boundaries, downloaded from gadm.org.
- I have taken your comments on board and have updated/tidied/condensed down the notebook a bit.
    - I also decided to also drop the section showing how to partition spatial points as I feel this wasn't necessary.

martinfleis:
Can you add `matplotlib` to doc/requirements.txt? The notebook doesn't render on RTD now https://dask-geopandas--154.org.readthedocs.build/en/154/guide/spatial-partitioning.html

tastatham:
> Can you add `matplotlib` to doc/requirements.txt? The notebook doesn't render on RTD now https://dask-geopandas--154.org.readthedocs.build/en/154/guide/spatial-partitioning.html

good point, I hadn't thought about that - fixed now.

martinfleis:
One thing to consider (I am not sure about it myself) is to mimic partition plotting spatialpandas has, which is a bit more explicit.

```py
import matplotlib.cm

fig, ax = plt.subplots(1,1, figsize=(12,6))
colors = matplotlib.cm.get_cmap("tab10").colors
for i, p in  enumerate(d_gdf.partitions):
    p.compute().plot(ax=ax, color=colors[i])
```
Before shuffle:
Screenshot 2022-02-24 at 15 02 30

After shuffle (with `tab20`):
Screenshot 2022-02-24 at 15 03 45

I like it because it is much easier to read than convex hulls but you should never do this with a `dask_geopandas.GeoDataFrame` of a reasonable size, so I am not sure if we should advertise it.

doc/source/guide/spatial-partitioning.ipynb
```diff
@@ -0,0 +1,383 @@
...
+    "\n",
+    "Dask-GeoPandas has implemented `spatial_shuffle` to repartition `Dask.GeoDataFrames`.\n",
+    "For those who are not familiar with Dask, a Dask DataFrame is internally split into many partitions, where each partition is a Pandas DataFrame.\n",
+    "Partitions are split vertically along the index; \n",
+    "\n",
+    "<img src=\"https://docs.dask.org/en/latest/_images/dask-dataframe.svg\" alt=\"drawing\" width=\"300\"/>\n",
+    "\n",
+    "\n",
+    "Partitioning a Dask DataFrame avoids element-wise operations, where every possible candidate in a query is evaluated.\n",
+    "By evaluating less candidates, we reduce memory overhead and the time taken to execute a query.\n",
```
martinfleis, line 18: What are you trying to say here? I am not sure it is correct.

tastatham:
> One thing to consider (I am not sure about it myself) is to mimic partition plotting spatialpandas has, which is a bit more explicit.
> 
> ```python
> import matplotlib.cm
> 
> fig, ax = plt.subplots(1,1, figsize=(12,6))
> colors = matplotlib.cm.get_cmap("tab10").colors
> for i, p in  enumerate(d_gdf.partitions):
>     p.compute().plot(ax=ax, color=colors[i])
> ```
> 
> Before shuffle: Screenshot 2022-02-24 at 15 02 30
> 
> After shuffle (with `tab20`): Screenshot 2022-02-24 at 15 03 45
> 
> I like it because it is much easier to read than convex hulls but you should never do this with a `dask_geopandas.GeoDataFrame` of a reasonable size, so I am not sure if we should advertise it.

I agree, that approach demonstrates more explicitly to the actual behaviour of spatial_shuffle.
If we demonstrated spatial shuffle using that approach, we should make it clear that users should not use matplotlib for plotting large Dask GeoDataFrames (e.g. millions of points) & if it's absolutely necessary, to use Datashader.

doc/source/guide/spatial-partitioning.ipynb
```diff
@@ -0,0 +1,383 @@
...
+    "us_cont.plot(facecolor=\"none\", linewidth=0.5, edgecolor=\"red\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e4509aa4-4036-4621-ad0c-2e87eef2340a",
+   "metadata": {},
+   "source": [
+    "We can then transform the GeoDataFrame to a Dask one, with a fixed number of aribitrary partitions - this does not yet spatially partition a Dask GeoDataFrame.\n",
+    "Note: we can also read (and export) parquet files in Dask-GeoPandas containing spatially partitioned data."
```
martinfleis, line 133: Use note admonition, like this (I think that the code suggestion would break this):
```
    ```{note} 
    We can also read (and export) parquet files in Dask-GeoPandas containing spatially partitioned data.
    ```
```

martinfleis:
@jorisvandenbossche what is your thought on this?

doc/source/guide/spatial-partitioning.ipynb
```diff
@@ -0,0 +1,383 @@
...
+    "\n",
+    "## Spatial sorting methods\n",
+    "spatial_shuffle supports three different partitioning methods, which provide an simple and inexpensive way of representing two-dimensional objects in one-dimensional space.\n",
+    "They are called using the `by` parameter;\n",
+    "1. Hilbert distance (default)\n",
+    "2. Morton distance\n",
+    "3. Geohash\n",
+    "\n",
+    "The first two methods are [space-filling curves](https://en.wikipedia.org/wiki/Space-filling_curve), which are lines that pass through every point in space, in a particular order and pass through points once only, so that each point lies a unique distance along the curve.\n",
+    "The range of both curves contain the entire 2-dimensional unit square, which means that the hilbert and morton curve can handle projected coordinates.\n",
```
martinfleis, line 197:
```suggestion
    "The range of both curves contain the entire 2-dimensional unit square, which means that the Hilbert and Morton curves can handle projected coordinates.\n",
```

doc/source/guide/spatial-partitioning.ipynb
```diff
@@ -0,0 +1,383 @@
...
+    "In general, we recommend the default Hilbert distance because it has better order-preserving behaviour.\n",
+    "This is because the curve holds a remarkable mathematical property, where the distance between two consecutive points along the curve is always equal to one.\n",
+    "By contrast, the Morton distance or [\"Z-order curve\"](https://en.wikipedia.org/wiki/Z-order_curve) does not hold this property and there are larger \"jumps\" in distances between two consecutive points.\n",
+    "Currently, calculated distances along the curves are based on the mid points of geometries.\n",
+    "Whilst this is inexpensive, mid points do not represent complex geometries well.\n",
+    "Future work will examine various methods for better representating complex geometries.\n",
+    "\n",
+    "The Geohash, not to be confused with [geohashing](https://geohashing.site/geohashing/Main_Page) is a hierarchical spatial data structure, which subdivides space into buckets of grid shape.\n",
```
martinfleis, line 205:
```suggestion
    "The Geohash, not to be confused with [geohashing](https://geohashing.site/geohashing/Main_Page), is a hierarchical spatial data structure, which subdivides space into buckets of grid shape.\n",
```

doc/source/guide/spatial-partitioning.ipynb
```diff
@@ -0,0 +1,383 @@
...
+    "Whilst the Geohash does allow binary searchers or spatial indexing, like rtree or quadtree, (a topic for later discussion & an enhancement for Dask-GeoPandas), Geohash is limited to encoding latitude and longitude coordinates.\n",
```
martinfleis, line 206:
```suggestion
    "Whilst the Geohash does allow binary searchers or spatial indexing, like rtree or quadtree, Geohash is limited to encoding latitude and longitude coordinates.\n",
```

doc/source/guide/spatial-partitioning.ipynb
```diff
@@ -0,0 +1,383 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e34f773c-0eec-4a78-bc78-66dd6c4ebd22",
+   "metadata": {},
+   "source": [
+    "## npartitions\n",
+    "The first parameter is the `npartitions` or number of partitions.\n",
+    "The default npartitions is the same number of partitions as the original Dask GeoDataFrame.\n",
+    "Whilst increasing the number of partitions will reduce the degree of spatial overlap between partitions, this is also a more expensive operation.\n",
```
martinfleis, line 260: > this is also a more expensive operation

But not by much right? This suggest significant difference while I don't think it will be so much more expensive.

doc/source/guide/spatial-partitioning.ipynb
```diff
@@ -0,0 +1,383 @@
...
+    "Knowing how many partitions to set is tricky because there is no single answer.\n",
+    "This depends not only on your data structure but also what you plan to do with your data.\n",
+    "Overall, there is a trade off between the up front and geometric operations cost when deciding how to spatially shuffle a Dask GeoDataFrame."
```
martinfleis, line 263: Can you rephrase this? I am not sure I am following what you're trying to say.

martinfleis, line 263: Can you rephrase this? I am not sure I am following what you're trying to say.

doc/source/guide/spatial-partitioning.ipynb
```diff
@@ -0,0 +1,383 @@
...
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b2d25cf2-891d-49b6-9b08-d4ebfabb6ed1",
+   "metadata": {},
+   "source": [
+    "## level\n",
+    "The `level` parameter stands for the precision of each partitioning method.\n",
```
martinfleis, line 340:
```suggestion
    "The `level` parameter represents the precision of each partitioning method.\n",
```

doc/source/guide/spatial-partitioning.ipynb
tastatham: Something like _Breaking your DataFrame into "too many" partitions results in [the scheduler incurring more overhead deciding where to compute each task](https://docs.dask.org/en/stable/dataframe-best-practices.html#repartition-to-reduce-overhead]._ ?

doc/source/guide/spatial-partitioning.ipynb
tastatham: With the us example provided there isn't much overhead going from 4 partitions to 10 partitions.
However, there is more overhead going from 10 to 100 partitions with natural earth low reg (~110ms vs 1s on my machine);

```python
import geopandas
import dask_geopandas

df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
ddf = dask_geopandas.from_geopandas(df, 4)

%timeit ddf.spatial_shuffle(npartitions=10)
%timeit ddf.spatial_shuffle(npartitions=100)
```
My understanding is that rearranging your DataFrame into smaller chunks incurrs more overhead than arranging it into fewer ones.
This should better link with Breaking your DataFrame into "too many" partitions results in [the scheduler incurring more overhead deciding where to compute each task
So it's a cost of creating a Dask GeoDataFrame vs cost of computation with each following task.

doc/source/guide/spatial-partitioning.ipynb
tastatham: Removed - I think this was copied from previous work.

doc/source/guide/spatial-partitioning.ipynb
martinfleis: What you see is mostly the scheduler overhead. With GeoDataFrames of a larger size, that is minimised.

```py
import geopandas
import dask_geopandas
import pandas

df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
ddf = dask_geopandas.from_geopandas(pandas.concat([df] * 1000), 4)

%timeit ddf.spatial_shuffle(npartitions=10)
%timeit ddf.spatial_shuffle(npartitions=100)

1.25 s ± 72 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.41 s ± 31.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

tastatham: Gotcha, I should have tested this with a larger dataset.
FYI, my results were;
```
1.54 s ± 36.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
2.63 s ± 237 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
which still isn't much - the most important thing to consider then is the scheduler incurring overhead, when deciding where to compute each subsequent task.
I'll update the notebook.

tastatham:
@martinfleis I think I have addressed each of the comments now, so it's ready for another/final review.

doc/source/guide/spatial-partitioning.ipynb
```diff
@@ -0,0 +1,717 @@
...
+   "source": [
+    "# Spatial partitioning in Dask-GeoPandas\n",
+    "\n",
+    "Dask-GeoPandas has implemented `spatial_shuffle` method to repartition `Dask.GeoDataFrames` geographically.\n",
+    "For those who are not familiar with Dask, a Dask DataFrame is internally split into many partitions, where each partition is a Pandas DataFrame.\n",
+    "Partitions are split vertically along the index; \n",
+    "\n",
+    "<img src=\"https://docs.dask.org/en/latest/_images/dask-dataframe.svg\" alt=\"drawing\" width=\"300\"/>\n",
+    "\n",
+    "There are numerous strategies that can be used to partition Dask DataFrames, which determines how the elements of a DataFrame are separated into each resulting partition.\n",
```
martinfleis:
```suggestion
    "There are numerous strategies that can be used to partition Dask DataFrames, which determine how the elements of a DataFrame are separated into each resulting partition.\n",
```

martinfleis (approved):
One tiny suggestion.

doc/source/guide/spatial-partitioning.ipynb
tastatham: I think you're right too

doc/source/guide/spatial-partitioning.ipynb
```diff
@@ -0,0 +1,717 @@
...
+    "\n",
+    "Dask-GeoPandas has implemented `spatial_shuffle` method to repartition `Dask.GeoDataFrames` geographically.\n",
+    "For those who are not familiar with Dask, a Dask DataFrame is internally split into many partitions, where each partition is a Pandas DataFrame.\n",
+    "Partitions are split vertically along the index; \n",
+    "\n",
+    "<img src=\"https://docs.dask.org/en/latest/_images/dask-dataframe.svg\" alt=\"drawing\" width=\"300\"/>\n",
+    "\n",
+    "There are numerous strategies that can be used to partition Dask DataFrames, which determine how the elements of a DataFrame are separated into each resulting partition.\n",
+    "Common strategies to partition Dask DataFrames include using a fixed number of partitions or partitions based on column values.\n",
+    "Dask-GeoPandas partitions GeoDataFrames by taking advantage of their spatial structure.\n",
```
jorisvandenbossche:
```suggestion
    "Dask-GeoPandas can partition GeoDataFrames by taking advantage of their spatial structure.\n",
```

(we don't necessarily do it, you need to explicitly do it, so "can" feels more correct)

doc/source/guide/spatial-partitioning.ipynb
```diff
@@ -0,0 +1,717 @@
...
...uRtI65VfKT0ocRN0kom1Z8FwXKc9FyvVq/nfheS4ce2VIkDhuxgyIpP1HYoy0POEd8Uq8YQWIi8q6ek5Zex1iXzKcALRKZGciROMjCHeMQ7nt3n5Cxe6vc4AieN6eHMKZMmSxeXJliyj6/9n78yDJruvME/zOvW/zPfaIjFwiE8jESgDcmRJFslRSSaJSUjWqqpep6rYe1thUz0zPTLXhjza0tbUZu9rGLG16qaopm5q2sipJJZU2ihQoiamFOwkIBEkQBEAQewKRe+wRvr/t3jN/PPdIj0iPCI/w50tEvJ8hEBnuz++7vr1773fP+U6nlwXmu6/bCcd40yAE79WCiwaR+3DW/O6Sr05frQUX1wM9UzDF0lzKaBaKS7J2Ejrmgam8FIK++e5K+T/uywm3OkXAMSUcU2KCgaofRQfXOixsfBhIw8UHxDwWeBQ3eApqOON/hhKhFM6/eRVjK+t4+6HzCOzOtJ8NdnHCTEvDYIQdlR7dmzEZib9ZOnjkutSqslytdR2dLH3Dr7vWNABcDUwsq/0s26OBRRAjBbrugmc0k6EAESUpR6fY4cHnbgD/pS1wC5Yij/hUHAHH5ZaVegCJgCXKAKwWQVjELAifGsvAMAy8vVDc0fo/IwV4B/vI3TCI6znJy2cc4b9T55kMvFdGhPv6LV34R4pF2iS1mkZwwyC92u3z6DeDiOxJI1pc3mLgzDdX66grjY8U+ivIBmt1kCEgeyhADyM10wEIyPqJEBwnDUsItV//114ho4JwFFMGSWwc5LK/k+fvYUIDYjff4O2Q1rDWSrsf07JLGYBMrZXOLsw/9/Hbrz1bsjLZJ5bf/dL9pdvztgoCAPCFYQBYAvDtgz+ThGNEq/qRiMAJRwpDq5ovjIOnTMdIKvQWJWu/YqYOWrW6I0KlENbrqNbbZ4QJIeDcFYnZtkzDsRtisWXB6kC4DQA7YNqi2jIRgtF8JP6a7ZcdOy3ta4YNTf2dxiiGaRmigqlMNjAFwvX2ieoC6DSbhomguFHkzhZwSyGPzjrGq+fT5luTllzPumH502Opb01YsgwAnmYDjawd57OPJVk7CZ2SPVlIvffuSnmgnSACsraBrG1AaUbFC1D2Qniq68DJIYAxQ2sYozLmeQZrfE+Ny4RdGFtZw0d+8GO8/dB5rE2O73l8ES4Uw8inw/payTxwCDYBmJAKs4ZCqgv
```
jorisvandenbossche: It's not super clear to me what is meant with this advice. Do you mean a workflow (eg a groupby) that might increase the number of partitions along the workflow?

jorisvandenbossche (commented):
Looks good!

doc/source/guide/spatial-partitioning.ipynb
tastatham: I'm not sure which part of the notebook you are referring to here?

jorisvandenbossche: Whoops, no idea how that got commented so out of place ..

jorisvandenbossche: Actually, so if you look in diff in the "Files changed" tab, then it appears at the correct location ..

tastatham: gotcha, the diff didn't load previously because of the notebook size.
I see what you mean about point number 3 - it's not very clear.
Yes, changing the number of partitions or even reverting back to GeoPandas could be useful after running an operation such as a  groupby.
I'll update the notebook later today.

jorisvandenbossche:
Thanks a lot @tastatham !

PR #156, opened 2022-02-11
BUG: ensure clip returns partitions as GeoSeries

### LINKED ISSUE #152: BUG: clip returns spatial_partitions as a GeoDataFrame
There's a bug in clip that returns new clipped `spatial_partitions` as a GeoDataFrame instead of a GeoSeries. Will post MRE later, just wanted to flag this to remember it.

martinfleis:
Closes #152 

I am not sure why we had `.to_frame("geometry")` in there but it is not needed. 

I have also changed one test to use real world data because that fixture returns LineStrings as partitions since all points are in line (I don't that that it is an issue by itself).

### DIFF
--- a/dask_geopandas/clip.py
+++ b/dask_geopandas/clip.py
@@ -23,7 +23,7 @@ def clip(gdf, mask, keep_geom_type=False):
         )
 
     new_spatial_partitions = geopandas.clip(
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):
Thanks!

PR #158, opened 2022-02-11
ENH: raise for empty geoms in sorting methods

### LINKED ISSUE #146: Handling of empty geometries in Hilbert/Morton distance methods
Currently the functionality "works" (i.e. doesn't break down) when empty geometries are present:

```
>>>s = geopandas.GeoSeries.from_wkt(["POINT (1 2)", "POINT (2 1)", "POLYGON EMPTY", "POINT (10 10)"])
>>> ds = dask_geopandas.from_geopandas(s, npartitions=1)
>>> ds.hilbert_distance().compute()
0     16691136
1      5509098
2            0
3    715827882
Name: hilbert_distance, dtype: uint32

```

but so empty geometries get filled in with 0. 

Is this OK? Or do we want to raise an error instead if there are empty geometries?

When using this in a context of spatial shuffling, this would mean that all empty ones end up in the same partition (which seems OK if for some reason you want to keep those empty geometries)

martinfleis:
Closes #146

Check for empty or missing geometry and raise `ValueError` instead of returning 0 in `hilbert_distance`, `morton_distance` and `geohash`.

### DIFF
--- a/dask_geopandas/geohash.py
+++ b/dask_geopandas/geohash.py
@@ -8,6 +8,8 @@ The vectorized implementation for quantization and bit interleaving is in turn b
 "Geohash in Golang Assembly" blog (https://mmcloughlin.com/posts/geohash-assembly).
 
 """
+import warnings
+
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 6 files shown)


dask_geopandas/geohash.py
```diff
@@ -31,6 +33,15 @@ def _geohash(gdf, as_string, precision):
     type : pandas.Series
         Series containing geohash
     """
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore", "GeoSeries.isna() previously returned True", UserWarning
+        )
```
martinfleis, line 39: I am also going to remove this from geopandas. It has been there for ages.

jorisvandenbossche: Yes, that's a good idea

jorisvandenbossche (approved):
Thanks, looks good!

Issue #66, opened 2021-06-28
ENH: change spatial_partitions from attribute into a property with getter/setter


jorisvandenbossche:
Currently the `spatial_partitions` is a plain attribute, meaning you can assign anything to it etc. We should probably turn it into a property, so we can control the actual getter/setter, and store the value itself in `_spatial_partitions`. Something like:

```python
@property
def spatial_partitions(self):
    return self._spatial_partitions

@spatial_partitions.setter
def spatial_partitions(self, value):
    # ...
    # ensure value is a GeoSeries, has the correct crs, has a default index, ...
    self._spatial_partitions = value
```

This would also easily allow to store the actual underlying value as something else as a GeoSeries (eg only bounding boxes, if we want that at some point), while keeping the public property returning a GeoSeries for convenience.

PR #159, opened 2022-02-11
Turn spatial_partitions into property + validate value in setter

jorisvandenbossche:
Closes https://github.com/geopandas/dask-geopandas/issues/66

@martinfleis this would have catched the bug in clip.py from https://github.com/geopandas/dask-geopandas/pull/156


martinfleis (approved):
Nice! Thanks!

PR #160, opened 2022-02-11
DOC: derive docstring for set_index

jorisvandenbossche:


martinfleis (approved):
Thanks! I think that we should eventually use custom decorator when inheriting docs from dask as it is a bit weird as it says

```
This docstring was copied from dask.dataframe.core.DataFrame.set_index.

Some inconsistencies with the Dask version may exist.
```

But for now it is good enough. We have the same issue with `to_parquet`.

jorisvandenbossche:
Indeed, that reads a bit strange.

PR #162, opened 2022-02-11
ENH: add GeoSeries.z and to_wkt, to_wkb on GeoSeries and GeoDataFrame

martinfleis:
Adding another batch of methods to get closer to the feature parity with GeoPandas (#130).


jorisvandenbossche (approved):
Thanks!

PR #163, opened 2022-02-12
Parquet IO: use dask pyarrow.dataset based engine instead of the legacy engine

jorisvandenbossche:
I noticed that we were actually subclassing the legacy engine from dask, and not the newer pyarrow.dataset based one. This doesn't change anything for the rest in our code (all tests pass), and this should also get rid of a bunch of deprecation warnings (from pyarrow, from using the legacy APIs)

It might not yet be available for older dask versions, in which case I can add a fallback to subclass the legacy engine in that case.

### DIFF
--- a/dask_geopandas/io/parquet.py
+++ b/dask_geopandas/io/parquet.py
@@ -8,9 +8,11 @@ from .arrow import GeoDatasetEngine, _get_partition_bounds, _update_meta_to_geod
 
 try:
     # pyarrow is imported here, but is an optional dependency
-    from dask.dataframe.io.parquet.arrow import ArrowEngine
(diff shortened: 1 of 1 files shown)


martinfleis:
> It might not yet be available for older dask versions, in which case I can add a fallback to subclass the legacy engine in that case.

Can we just bump requirements? We currently require a version from 6 Jun 2020.

jorisvandenbossche:
The current failure is actually not because of this PR, but so the Feather PR where I started using `from dask.layers import DataFrameIOLayer`. 

Checking this new class, it seems to be introduced in https://github.com/dask/dask/pull/7415, which corresponds to dask 2021.05.0

martinfleis (approved):

PR #164, opened 2022-02-14
DEP: bump dask minimum required version to 2021.06.0

jorisvandenbossche:
See https://github.com/geopandas/dask-geopandas/pull/163#issuecomment-1039452906

With the latest changes of https://github.com/geopandas/dask-geopandas/pull/91, we accidentally started to require dask 2021.5.0. So one option to deal with this is to simply bump the minimum required version (I directly bumped to 2021.06.0, because we already disallow 2021.05.1)

### DIFF
--- a/continuous_integration/envs/37-latest.yaml
+++ b/continuous_integration/envs/37-latest.yaml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   # required dependencies
   - python=3.7
-  - dask=2.18
+  - dask=2021.06.0
   - distributed
   - geopandas
   - fiona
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
I think this is probably fine (given the early stage of this package), but just to be explicit: this is only for the feather IO, so it is also easy to just give an error specifically for that in case dask is too old.

jorisvandenbossche:
Ah, now it fails again with the fiona error ..

martinfleis (approved):
I'd merge this and wait for fiona to be resolved elsewhere. We can potentially switch back to conda or remove defaults channel altogether if it proves to be a long-term issue.

jorisvandenbossche:
OK, pinning GDAL to force it to use the older version from conda-forge seems to have worked, and all is green now!

PR #166, opened 2022-02-18
DOC: add pypi badge

raybellwaves:
I find this badge helpful as a quick view of the version number.

You can see how it looks here: https://github.com/raybellwaves/dask-geopandas/tree/patch-1

Screen Shot 2022-02-18 at 2 25 01 PM


jorisvandenbossche (approved):
Thanks!

PR #173, opened 2022-02-19
Packaging updates: add pyproject.toml, update setup.py, move tests inside package

jorisvandenbossche:


setup.py
```diff
@@ -40,8 +49,9 @@
         "Source": "https://github.com/geopandas/dask-geopandas/",
         "Issue Tracker": "https://github.com/geopandas/dask-geopandas/issues",
     },
-    packages=find_packages(),
+    packages=find_packages(exclude=("tests", "tests.*")),
```
jorisvandenbossche, line 52: I was checking the wheels we created, and noticed that they have a top-level `tests/` directory (which gets installed as a "tests" package in site-packages), which reminded me of https://twitter.com/ocefpaf/status/1466456847125200896 :)

setup.py
```diff
@@ -26,11 +33,13 @@
...
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
     ],
     keywords="dask geopandas spatial distributed cluster",
-    description="GeoPandas objects backed with Dask",
+    description="Parallel GeoPandas with Dask",
```
jorisvandenbossche, line 42: I updated this with the "tagline" we have in the README / github description. Do we find this a good one? Can also update it everywhere to yet something else (I am wondering if "parallel" is clear enough that it is also "distributed")

.github/workflows/release_to_pypi.yml
```diff
@@ -1,13 +1,19 @@
-name: Publish dask-geopandas to PyPI / GitHub
+name: Build and publish dask-geopandas to PyPI / GitHub
 
 on:
   push:
     tags:
       - "v*"
+  pull_request:  # also build on PRs touching this file
+    paths:
+      - ".github/workflows/releaseto_pypi.yml"
```
martinfleis, line 9:
```suggestion
      - ".github/workflows/release_to_pypi.yml"
```

setup.py
martinfleis: I would move `tests` to `dask-geopandas` folder so we ship them with the package. You can then run them from an installed environment which is useful e.g. when verifying if you docker container is properly built.

setup.py
martinfleis: Dask itself uses `parallel` primarily, so I am fine with that.

setup.py
jorisvandenbossche: Fine for me as well, moved them inside

martinfleis (approved):
Looks okay to me. Let's hope it will do the job once triggered.

jorisvandenbossche:
Yeah, I think the main uncertainty is the re-downloading of the artifacts. We can always do another alpha first to ensure everything is working. 

Now, it's maybe also possible to avoid this re-downloading, if I make it again a single workflow. I have to check if it is easy to skip all subsequent steps with an `if` clause (instead of just skipping one step)

PR #174, opened 2022-02-19
CI: add test env with no optional dependencies

jorisvandenbossche:
Now we have both pyarrow and pyogrio as optional dependencies for IO, we should ensure we don't accidentally make them required dependencies by importing them unconditionally. Copying the testing approach from geopandas.


martinfleis (approved):
Thanks!

PR #176, opened 2022-02-24
DOC: Update spatial_shuffle level using default level 16

tastatham:
The current docstring for `spatial_shuffle` incorrectly asserts that the the default level is 15.
This should be 16.


martinfleis (approved):
Good catch!

PR #178, opened 2022-02-25
DOC: move doc/source/docs/reference to doc/source/reference

jorisvandenbossche:
@martinfleis this might be a left-over from the initial doc structure, or is this intentional? I think our full `doc` directory is the "docs", and since the "guide" is also directly inside the `doc/source/` directory, seems logical to do the same for "reference"?


martinfleis:
That is intentional. Since we are inheriting a lot of our docstrings from GeoPandas, we need to keep the same URL to API to ensure the links work. Like this one to the figure https://dask-geopandas--178.org.readthedocs.build/en/178/reference/api/dask_geopandas.GeoSeries.touches.html

jorisvandenbossche:
Ah, now I remember again that I already asked about this before :)

Issue #76, opened 2021-07-09
BUG: cx indexer doesn't work when spatial_partitions are not defined


jorisvandenbossche:
Inside the CoordinateIndexer (for `.cx[...]`), we access `xmin, ymin, xmax, ymax = obj.spatial_partitions.total_bounds`, which only works if `spatial_partitions` is not None. We should another code path for if it is not set.

PR #181, opened 2022-03-09
Raise proper error message in .cx indexer if no spatial_partitions are set

jorisvandenbossche:
Related to https://github.com/geopandas/dask-geopandas/issues/76. Doesn't yet implement it when `spatial_partitions is None`, but should at least give a better error message (about an attribute error about None not having total_bounds)


jorisvandenbossche:
The Mac failure is a known issue on the conda-forge side (https://github.com/conda-forge/fontconfig-feedstock/pull/53#issuecomment-1062456875)

martinfleis (approved):
Thanks!

Issue #190, opened 2022-05-19
Empty partitions cause all columns to be promoted to float


fbunt:
If one of the partitions comes back empty then `dask_geopandas` promotes all columns to float. I checked if the same thing happens with `dask.dataframe` and it does NOT. This seems to be an issue with `dask_geodataframe`, as far as I can tell. Below is an extremely simplified version of my use case. In it, I'm processing chunks of a dask array and outputting a GeoDataFrame based on the contents. If all of the values in a chunk are missing (0), then I return an empty dataframe.

```python
import dask
import dask.array as da
import dask.dataframe as dd
import geopandas as gpd
import dask_geopandas as dgpd
from shapely.geometry import Point


@dask.delayed
def chunk_to_df(x):
    values = []
    rows = []
    cols = []
    points = []
    for i in range(x.shape[0]):
        for j in range(x.shape[1]):
            if x[i, j] == 0:
                continue
            values.append(x[i, j])
            rows.append(i)
            cols.append(i)
            points.append(Point(float(j), float(i)))
    return gpd.GeoDataFrame(
        {"value": values, "row": rows, "col": cols, "geometry": points}
    )


data = da.from_array(
    np.array(
        [
            [0, 0, 1, 2],
            [0, 0, 2, 2],
            [0, 0, 1, 0],
            [0, 1, 3, 0],
        ]
    ),
    chunks=(2, 2),
)
meta = gpd.GeoDataFrame(
    {
        "value": [1],
        "row": [0],
        "col": [0],
        "geometry": Point(1.0, 1.0),
    }
)
ddf = dd.concat(
    [
        dd.from_delayed(chunk_to_df(chunk), meta=meta)
        for chunk in data.to_delayed().ravel()
    ]
)

print(ddf)
print(ddf.partitions[0].compute())
print(ddf.partitions[1].compute())
print(ddf.compute())
print(ddf.compute().dtypes)
```
### Actual Output

```python
Dask GeoDataFrame Structure:
               value    row    col  geometry
npartitions=4
               int64  int64  int64  geometry
                 ...    ...    ...       ...
                 ...    ...    ...       ...
                 ...    ...    ...       ...
                 ...    ...    ...       ...
Dask Name: concat, 16 tasks
Empty GeoDataFrame
Columns: [value, row, col, geometry]
Index: []
   value  row  col                 geometry
0      1    0    0  POINT (0.00000 0.00000)
1      2    0    0  POINT (1.00000 0.00000)
[...]

PR #191, opened 2022-05-21
Remove custom __dask_postcompute__

jorisvandenbossche:
Closes https://github.com/geopandas/dask-geopandas/issues/190

I _think_ our custom `__dask_postcompute__` shouldn't be needed, because `pd.concat` should already preserve the CRS of the GeoDataFrames.

### DIFF
--- a/dask_geopandas/tests/test_core.py
+++ b/dask_geopandas/tests/test_core.py
@@ -833,3 +833,19 @@ def test_to_dask_dataframe(geodf_points_crs):
     assert isinstance(result, pd.DataFrame) and not isinstance(
         result, geopandas.GeoDataFrame
     )
+
+
+def test_compute_empty_partitions():
+    # https://github.com/geopandas/dask-geopandas/issues/190 - ensure to skip
+    # empty partitions when concatting the computed results
+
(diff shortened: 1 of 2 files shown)


dask_geopandas/tests/test_core.py
```diff
@@ -833,3 +833,19 @@ def test_to_dask_dataframe(geodf_points_crs):
     assert isinstance(result, pd.DataFrame) and not isinstance(
         result, geopandas.GeoDataFrame
     )
+
+
+def test_compute_empty_partitions():
+    # https://github.com/geopandas/dask-geopandas/issues/190 - ensure to skip
+    # empty partitions when concatting the computed results
```
jorisvandenbossche, line 840: This test is based on the report from https://github.com/geopandas/dask-geopandas/issues/190, but it is maybe not a fully "correct" test, in that it (accidentally, because of pandas defaulting to float64 for empty data) creates a somewhat "incorrect" dask GeoDataFrame (the first empty partition has dtypes not matching `meta`). 

But it _is_ a test that currently fails on main, and passes with this PR.

Issue #198, opened 2022-06-14
dask 2022.6.0 causes ArrowTypeError in to_parquet


fbunt:
With the latest version of dask (2022.6.0), calling `to_parquet()` on a dask_geopandas dataframe throws `ArrowTypeError: Did not pass numpy.dtype object`. I tried replicating with just a `dask.dataframe.DataFrame` but no error was thrown. It only seems to happen for me with dask_geopandas.

## Example

```python
import pandas as pd
import geopandas as gpd
import dask_geopandas as dgpd

dft = pd.util.testing.makeDataFrame()
dft["geometry"] = gpd.points_from_xy(dft.A, dft.B)
df = gpd.GeoDataFrame(dft)
df = dgpd.from_geopandas(df, npartitions=1)
df.to_parquet("mydf.parquet")
```

```python
---------------------------------------------------------------------------
ArrowTypeError                            Traceback (most recent call last)
Input In [47], in <cell line: 1>()
----> 1 df.to_parquet("mydf.parquet")

File ~/anaconda3/envs/rstools2/lib/python3.8/site-packages/dask/dataframe/core.py:5045, in DataFrame.to_parquet(self, path, *args, **kwargs)
   5042 """See dd.to_parquet docstring for more information"""
   5043 from dask.dataframe.io import to_parquet
-> 5045 return to_parquet(self, path, *args, **kwargs)

File ~/anaconda3/envs/rstools2/lib/python3.8/site-packages/dask/dataframe/io/parquet/core.py:844, in to_parquet(df, path, engine, compression, write_index, append, overwrite, ignore_divisions, partition_on, storage_options, custom_metadata, write_metadata_file, compute, compute_kwargs, schema, name_function, **kwargs)
    835     raise ValueError(
    836         "User-defined key/value metadata (custom_metadata) can not "
    837         "contain a b'pandas' key.  This key is reserved by Pandas, "
    838         "and overwriting the corresponding value can render the "
    839         "entire dataset unreadable."
    840     )
    842 # Engine-specific initialization steps to write the dataset.
    843 # Possibly create parquet metadata, and load existing stuff if appending
--> 844 i_offset, fmd, metadata_file_exists, extra_write_kwargs = engine.initialize_write(
    845     df,
    846     fs,
    847     path,
    848     append=append,
    849     ignore_divisions=ignore_divisions,
    850     partition_on=partition_on,
    851     division_info=division_info,
    852     index_cols=index_cols,
    853     schema=schema,
[...]

PR #199, opened 2022-06-17
Compat with dask 2022.06.0: fix schema inference in to_parquet

jorisvandenbossche:
Closes https://github.com/geopandas/dask-geopandas/issues/198


jorisvandenbossche:
> Shall we cut a patch release?

Yes, was thinking the same!

Issue #194, opened 2022-06-01
Slow `read_parquet` with many files.


TomAugspurger:
I'm looking into the performance of `dask_geopandas.read_parquet` when we have a Parquet dataset with many files. Dropping some notes here.

## Context

I'm working on Azure Blob Storage, but this should apply to S3 and GCS as well.

The structure is

```
/path/to/root.parquet
    /region=A
      - part.0.parquet
      - part.1.parquet
    /region=B
      - part.0.parquet
      - part.1.parquet
      - ...
...
```

There are 206 "regions" and 3,679 parts in total.

At the moment, there's no `_metadata` or `_global_metadata` file present. I could probably add one if that's helpful.

## Minimal Example

```python
import dask_geopandas
import requests
import adlfs

token = requests.get("https://planetarycomputer-staging.microsoft.com/api/sas/v1/token/bingmlbuildings/footprints").json()["token"]

storage_options=dict(account_name="bingmlbuildings", credential=token)

%time df = dask_geopandas.read_parquet("az://footprints/geo/2022-05-25/ml-buildings.parquet", storage_options=storage_options)
````

From the same Azure region as the data, that takes about 3 minutes to complete.

## Reasons

I think most of the time is being spent in https://github.com/geopandas/dask-geopandas/blob/16561258318d2d7a2ae24c06741019050b3645f2/dask_geopandas/io/parquet.py#L26-L34 (need to confirm this)

What are the consequences of not providing that metadata? That we end up with incorrect / not `spatial_partitions` I suspect? Depending on the use case, unset spatial divisions might be OK. Would it be acceptable to disable those boundary reads with a flag in `read_parquet`?

## Relevant links:

- https://github.com/dask/dask/issues/8058

jorisvandenbossche:
Reading this metadata is indeed only for populating the `spatial_partitions`

Adding an option to not load the spatial partitioning information upon construction of the dask objects sounds as a good idea.

On the actual implementation, I am wondering if this can actually be improved. In the parent ArrowDatasetEngine class from dask, we actually do have a pyarrow dataset object and fragments, but after `read_metadata`  dask typically extracts the file path for each fragment. If we would have more control over this, we could access the schema directly from the fragment. Although I am not sure how much benefit this will give, since accessing the schema can still invoke IO at that point to read the physical schema from each file (very similar as how gathering the statistics in dask's read_parquet gives a lot of IO when constructing the object, I _think_ this is no longer done by default?). 

[...]

PR #203, opened 2022-06-20
Optionally skip spatial bounds in read_parquet

TomAugspurger:
Adds a new `gather_spatial_partitions` keyword to `read_parquet`
to disable opening each file to get its spatial bounds. The name was
chosen to mimic dask's `gather_statistics` keyword.

Also adds a small docs section (I didn't see an easy way to
insert a snippet in the docstring).

Closes https://github.com/geopandas/dask-geopandas/issues/194.

---

One note of hesitation: I think Dask mid-transition for handling how it reads metadata. I'm wondering whether we should just rely on the behavior of dask's `gather_statistics` keyword. IIUC, both it and this new `gather_spatial_partitions` control whether there's a *per-file* operation in `read_parquet`.

Maybe @jcrist or @rjzamora have a recommendation on whether adding a new keyword here is going against where Dask is headed.

### DIFF
--- a/dask_geopandas/io/parquet.py
+++ b/dask_geopandas/io/parquet.py
@@ -50,13 +50,16 @@ class GeoArrowEngine(GeoDatasetEngine, DaskArrowDatasetEngine):
     def read_metadata(cls, fs, paths, **kwargs):
         meta, stats, parts, index = super().read_metadata(fs, paths, **kwargs)
 
-        # get spatial partitions if available
-        regions = geopandas.GeoSeries(
-            [_get_partition_bounds_parquet(part, fs) for part in parts], crs=meta.crs
-        )
-        if regions.notna().all():
-            # a bit hacky, but this allows us to get this passed through
-            meta.attrs["spatial_partitions"] = regions
+        gather_spatial_partitions = kwargs.pop("gather_spatial_partitions", True)
+
+        if gather_spatial_partitions:
+            regions = geopandas.GeoSeries(
+                [_get_partition_bounds_parquet(part, fs) for part in parts],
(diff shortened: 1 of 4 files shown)


TomAugspurger:
cc @jorisvandenbossche.

jorisvandenbossche:
The `gather_statistics` keyword is deprecated now in dask, in favor of keywords for the explicit end behaviour you want (eg `calculate_divisions=True`, or `split_row_groups=True`, for which in both cases the statistics (or more generally parquet file metadata) needs to be read).  
So I _think_ adding a keyword like this specifically for controlling the spatial partitions seems in line with the latest changes in dask.

TomAugspurger:
OK, thanks. In that case, I think a `calculate_spatial_divisions` or `calculate_spatial_partitions` keyword is appropriate, mirroring `calculate_divisions` (https://docs.dask.org/en/stable/dataframe-parquet.html#calculating-divisions).

For now I'll go with `calculate_spatial_partitions`.

martinfleis:
We talked about that a bit and `calculate` is not necessarily a right word as dask is not _calculating_ but _gathering_ bounds that are stored in the parquet meta. Calculate imposes that dask will read all geometries and get total_bounds of those for each partition, which is not the case.

TomAugspurger:
OK, reverted to go back to `gather_spatial_partitions`.

martinfleis:
Hey, I'll have a look later tonight and we can even cut 0.2.0. We already talked about that last week with @jorisvandenbossche.

martinfleis:
I'll go ahead and merge this, then we should ideally get #205 in and then can cut 0.2.0.

PR #205, opened 2022-06-23
Don't put GeoSeries in `map_partitions` kwarg

### LINKED ISSUE #197: Creating dask geodataframe from_dask_dataframe deadlocks
I'm running into issues when I create a dask geodataframe from a regular dask dataframe. After some debugging it looks like the `.finalize()` task is causing issues. 

The local reproducer below shows that the `finalize` task is created. This is not a problem when the data fits in memory. T

However, when working at scale, this `finalize` call causes workers to run out of memory. These workers are not killed but become unresponsive, causing a virtual 'deadlock' where the `finalize` task gets endlessly shipped around workers.

### Local Reproducer (does not hang because data fits in memory)
```python
import dask.dataframe as dd
from distributed import Client

client = Client(n_workers=8)

# load some data
ddf = dd.read_csv(
    "s3://nyc-tlc/csv_backup/yellow_tripdata_2012-01.csv",
)

# subset for faster iteration
ddf = ddf.partitions[0:5]

# convert to dask geodataframe
import dask_geopandas
ddf = dask_geopandas.from_dask_dataframe(
    ddf,
    geometry=dask_geopandas.points_from_xy(ddf, "pickup_longitude", "pickup_latitude"),
)

ddf.head()


### Cloud-Based Reproducer (hangs because data does not fit into memory)

```python
import dask.dataframe as dd

ddf = dd.read_parquet(
    "s3://coiled-datasets/dask-book/nyc-tlc/2009-2013/*",
    engine="pyarrow",
    storage_options={"anon": True},
)

import dask_geopandas

ddf = dask_geopandas.from_dask_dataframe(
    ddf,
    geometry=dask_geopandas.points_from_xy(ddf, "pickup_longitude", "pickup_latitude"),
)

ddf.head()
```

image

ian-r-rose:
Workaround for the fact that we can't include a partitioned `_Frame` in the keyword arguments for `map_partitions without it being concatenated on a single worker.

Fixes #197

### DIFF
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -778,6 +778,15 @@ def from_dask_dataframe(df, geometry=None):
         the values will be set as 'geometry' column on the GeoDataFrame.
 
     """
+    # If the geometry column is already a partitioned `_Frame`, we can't refer to
+    # it via a keyword-argument due to https://github.com/dask/dask/issues/8308.
+    # Instead, we assign the geometry column using regular dataframe operations,
+    # then refer to that column by name in `map_partitions`.
+    if isinstance(geometry, _Frame):
+        name = geometry.name or "geometry"
+        return df.assign(**{name: geometry}).map_partitions(
+            geopandas.GeoDataFrame, geometry=name
+        )
     return df.map_partitions(geopandas.GeoDataFrame, geometry=geometry)
 
 


dask_geopandas/tests/test_core.py
```diff
@@ -390,6 +390,12 @@ def test_from_dask_dataframe_with_dask_geoseries():
     dask_obj = dask_geopandas.from_dask_dataframe(
         dask_obj, geometry=dask_geopandas.points_from_xy(dask_obj, "x", "y")
     )
+    # Check that the geometry isn't concatenated and embedded a second time in
```
ian-r-rose, line 393: This test stinks, but I wasn't really able to think of a better way to do this.

dask_geopandas/core.py
```diff
@@ -778,6 +778,15 @@ def from_dask_dataframe(df, geometry=None):
         the values will be set as 'geometry' column on the GeoDataFrame.
 
     """
+    # If the geometry column is already a partitioned `_Frame`, we can't refer to
+    # it via a keyword-argument due to https://github.com/dask/dask/issues/8308.
+    # Instead, we assign the geometry column using regular dataframe operations,
+    # then refer to that column by name in `map_partitions`.
+    if isinstance(geometry, dd.core._Frame):
+        name = geometry.name or "geometry"
```
TomAugspurger, line 786: This will behave incorrectly for falsey names like `0` or `False`. 

```suggestion
        if geometry.name is None:
            name = geometry.name
        else:
            name = "geometry"
```

dask_geopandas/core.py
```diff
@@ -778,6 +778,15 @@ def from_dask_dataframe(df, geometry=None):
         the values will be set as 'geometry' column on the GeoDataFrame.
 
     """
+    # If the geometry column is already a partitioned `_Frame`, we can't refer to
+    # it via a keyword-argument due to https://github.com/dask/dask/issues/8308.
+    # Instead, we assign the geometry column using regular dataframe operations,
+    # then refer to that column by name in `map_partitions`.
+    if isinstance(geometry, dd.core._Frame):
```
TomAugspurger, line 785: Should this just be `dd.Series`? I don't think passing a `dd.DataFrame` here would be valid.

ian-r-rose: Good point

TomAugspurger (approved):
Looks good. Thanks Ian!

martinfleis:
Thanks @ian-r-rose and @TomAugspurger!

PR #216, opened 2022-08-27
BUG: convert spatial partitions to geoseries in sjoin result

slumnitz:
* make sure 'core.py' does not skip self.spatial_partitions quality check by removing hidden function bypass
* add conversion to geoseries in `sjoin.py'


martinfleis (approved):
Nice catch! Thanks @slumnitz!

jorisvandenbossche:
Thanks!

Issue #40, opened 2021-03-12
Overlapping computations


martinfleis:
If you want to measure anything which has to do with the spatial relationship of features, you eventually find an issue of needing to go across chunk boundaries (assuming spatially chunked ddf). A typical example is a spatial lag, i.e. the mean of values on neighbouring (touching) features. 

The way raster analysis deals with it is using `dask.array.map_overlap`. That essentially copies a bit of neighbouring data to each chunk to create overlaps so you can do your spatial lag fully within a chunk. See https://docs.dask.org/en/latest/array-overlap.html

I believe that we need the analogy of `map_overlap` for vector data. It is naturally a significantly more complex issue since we do not know how the data look like (it is not a grid). But I believe it is doable and could be a massive game-changer. For example, I would be able to base 90% of momepy on dask-geopandas.

The trick is to define which features should be overlapping. For that, you need to know how far you have to go for each particular operation but we can specify:

- a distance threshold (everything within _n_ meters from the chunk boundary)
- topological threshold (everything within _n_ steps of contiguity)

I have actually already tested this approach with topological threshold, with custom single-core functions and it works well.

We obviously first need spatial re-chunking and spatial indexing, but this is something I'd like to put on a roadmap (maybe for GSoC?).

PR #217, opened 2022-08-27
[ENH] Overlay functionality with Dask-geopandas

slumnitz:
* add overlay.py script to codebase
* add overlay to __init__.py

To-Do
- [ ] add tests
- [ ] adding keywords

### DIFF
--- a/dask_geopandas/__init__.py
+++ b/dask_geopandas/__init__.py
@@ -32,5 +32,5 @@ __all__ = [
     "to_parquet",
     "clip",
     "sjoin",
-    "overlay"
+    "overlay",
 ]
--- a/dask_geopandas/__init__.py
+++ b/dask_geopandas/__init__.py
@@ -13,6 +13,7 @@ from .io.file import read_file
 from .io.parquet import read_parquet, to_parquet
 from .io.arrow import read_feather, to_feather
 from .sjoin import sjoin
+from .overlay import overlay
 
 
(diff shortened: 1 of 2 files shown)


martinfleis:
Thanks for looking into this. Seeing that this is essentially a mirror of sjoin inner implementation we have here, I am a bit afraid it won't be that easy. 

Take an example of `difference` and assume you have a polygon in `df1` which intersects with two polygons in `df2`, each of which needs to be clipped out from the original. The issue you may face is when one of these two polygons is in one partition while the other is in another one. What happens in this PR is that you will end up with two versions of our original polygon from `df1`, one that is a result of difference with one partition and the other that comes from the operation with the other partition. So we have a duplicated entry where both of them are actually wrong.

We face very similar issue in left `sjoin`, that is why only inner is implemented at the moment.

For overlay, we would need some form of very smart overlapping computation (#40) and I am not fully sure how would that work at the moment.

keewis:
I wonder if it would not be better to solve this using the map-reduce pattern? For computing differences, 
the idea would be to first compute the element-wise differences, then group by the geometries we subtract from and reduce using intersection. The groupby-reduce can be implemented using a tree-reduce (the picture is from the [`flox` documentation](https://flox.readthedocs.io/en/latest/implementation.html#method-map-reduce)):



That way, we can avoid increasing the partition sizes to implement `map_overlap`, which could make some of the partitions bigger-than-memory.

Issue #225, opened 2022-12-02
Bug reading parquet files with `dask==2022.12.0`


the-matt-morris:
Hello, thanks for all the great work on this library!

I've been using this library as a dependency in one of my projects and just updated `dask` to `2022.12.0`.  With that change, I'm now receiving an error when I attempt to read a parquet file in from s3.  I have confirmed that this error exists with `dask==2022.12.0` but does not occur with `dask==2022.11.1`.  Using `dask_geopandas==0.2.0` on ubuntu 20.04.

Error:

```
>>> dgpd.read_parquet('s3://<MY_BUCKET>/boundaries.parquet')
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/dask/backends.py", line 125, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/dask/dataframe/io/parquet/core.py", line 479, in read_parquet
    read_metadata_result = engine.read_metadata(
  File "/usr/local/lib/python3.9/site-packages/dask_geopandas/io/parquet.py", line 51, in read_metadata
    meta, stats, parts, index = super().read_metadata(fs, paths, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/dask/dataframe/io/parquet/arrow.py", line 372, in read_metadata
    meta = cls._create_dd_meta(dataset_info, use_nullable_dtypes)
TypeError: _create_dd_meta() takes 2 positional arguments but 3 were given

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/site-packages/dask_geopandas/io/parquet.py", line 102, in read_parquet
    result = dd.read_parquet(*args, engine=GeoArrowEngine, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/dask/backends.py", line 127, in wrapper
    raise type(e)(
TypeError: An error occurred while calling the read_parquet method registered to the pandas backend.
Original Message: _create_dd_meta() takes 2 positional arguments but 3 were given
```

[...]

PR #228, opened 2023-01-06
CI: update versions in testing

jorisvandenbossche:
Some long overdue updates, since we were not yet testing with python 3.10 and 3.11


jorisvandenbossche:
So with those changes, the actual builds that use latest dask (311-latest and 311-dev) are still failing because of https://github.com/geopandas/dask-geopandas/issues/225 (and that can be handled in a separate PR). The other builds use older dask versions that pass at the moment.

PR #229, opened 2023-01-06
COMPAT: specify shuffle=False in dissolve

jorisvandenbossche:
On top of https://github.com/geopandas/dask-geopandas/pull/228 (will rebase this once the other is merged)

Starting with dask 2022.9.1, groupby started to use shuffle by default with `split_out>1`, see https://github.com/dask/dask/pull/9453

But doing that currently doesn't work (at least with the test case). It comes down to `_meta_nonempty` failing to be determined, an issue with renamed columns and then the GeoDataFrame constructor fails with that:

https://github.com/geopandas/dask-geopandas/blob/bd355404dad23c8a1a5876834f569d54d85fc626/dask_geopandas/backends.py#L60-L63

First, it raises "The CRS attribute of a GeoDataFrame without an active geometry column is not defined. Use GeoDataFrame.set_geometry to set the active geometry column.", and after removing `crs=x.crs`, it raises "ValueError: Unknown column geometry" because the groupby shuffle renamed the columns

(I remember we ran into similar issues before, but not fully sure anymore if we solved that / where this discussion is)

### DIFF
--- a/continuous_integration/envs/310-latest.yaml
+++ b/continuous_integration/envs/310-latest.yaml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   # required dependencies
   - python=3.10
-  - dask=2022.06.0
+  - dask=2022.10.0
   - distributed
   - geopandas
   - pygeos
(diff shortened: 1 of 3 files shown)


martinfleis (dismissed):
Thanks! Feel free to merge #228 as well once happy, looks okay to me.

jorisvandenbossche:
Ah, indeed, the `shuffle` keyword was added in https://github.com/dask/dask/pull/9302 with dask 2022.8.1

PR #230, opened 2023-01-06
BUG: fix `read_parquet` with `dask>=2022.12.0`

### LINKED ISSUE #225: Bug reading parquet files with `dask==2022.12.0`

the-matt-morris:
closes #225 

Updates method signature of the following so that they are compatible with all versions of `dask`, including `dask>=2022.12.0`:

* `dask_geopandas.io.arrow.ArrowDatasetEngine._arrow_table_to_pandas`
* `dask_geopandas.io.arrow.GeoDatasetEngine._arrow_table_to_pandas`
* `dask_geopandas.io.parquet.GeoArrowEngine._create_dd_meta`

It's been a little while since I played around with this, but I had written previously that [these lines in dask](https://github.com/dask/dask/blob/f463fcf28ff6404cc04f478b73704e35664bb5f0/dask/dataframe/io/parquet/arrow.py#L1073-L1078) are problematic, as it calls `_arrow_table_to_pandas`, where it wasn't called prior to `dask==2022.12.0`.

At the time, I thought we may need an update to `GeoDatasetEngine._arrow_table_to_pandas` to handle this, but I feel like someone with a better grasp on this library's internals may have a better idea.  Either way, it is known that this PR in its current state is an attempt to solve the problem, but it doesn't succeed entirely in doing that, as you'll see with one of the unit tests failing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 22.10.0
+    rev: 22.12.0
     hooks:
       - id: black
         language_version: python3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,11 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 22.10.0
     hooks:
       - id: black
         language_version: python3
-  - repo: https://gitlab.com/pycqa/flake8
-    rev: 4.0.1
+  - repo: https://github.com/pycqa/flake8
+    rev: 6.0.0
     hooks:
       - id: flake8
         language_version: python3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
       - id: black
         language_version: python3
   - repo: https://github.com/pycqa/flake8
-    rev: 6.0.0
+    rev: 5.0.4
     hooks:
       - id: flake8
         language_version: python3
--- a/dask_geopandas/io/arrow.py
+++ b/dask_geopandas/io/arrow.py
@@ -126,51 +126,35 @@ class ArrowDatasetEngine:
 
         return fragments, meta, schema, filter
 
-    if DASK_2022_12_0_PLUS:
+    @classmethod
+    def _arrow_table_to_pandas(
+        cls,
+        arrow_table: "pyarrow.Table",
+        categories,
+        **kwargs,
+    ) -> pd.DataFrame:
 
-        @classmethod
-        def _arrow_table_to_pandas(
-            cls,
-            arrow_table: "pyarrow.Table",
-            categories,
-            use_nullable_dtypes=False,
-            **kwargs,
-        ) -> pd.DataFrame:
+        _kwargs = kwargs.get("arrow_to_pandas", {})
+        _kwargs.update({"use_threads": False, "ignore_metadata": False})
+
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
> Either way, it is known that this PR in its current state is an attempt to solve the problem, but it doesn't succeed entirely in doing that, as you'll see with one of the unit tests failing.

There are two failing tests at the moment. One will be fixed by https://github.com/geopandas/dask-geopandas/pull/229 (and is unrelated to the parquet IO), and the other is a case that should error but is giving a wrong error message. But that indeed seems to be related to the lines you link to (`_arrow_table_to_pandas` is being called with an empty table, and that is not working with our implementation). I will take a look at that.

jorisvandenbossche:
And thanks for opening the PR!

the-matt-morris:
You're right, I was confused as to why that other test was failing, glad there's a PR for that one already.  Thanks for checking it out!

jorisvandenbossche:
We can probably further improve this and make the integration more robust, but let's already merge this so we can have green CI and have this running with latest dask.

Thanks again @the-matt-morris!

Issue #226, opened 2022-12-16
Method directly calls PyGEOS function, but GeoPandas is deprecating PyGEOS for Shapely


gtmaskall:
New GeoPandas issues warning about PyGEOS:
```
/tmp/ipykernel_24453/103584512.py:4: UserWarning: Shapely 2.0 is installed, but because PyGEOS is also installed, GeoPandas will still use PyGEOS by default for now. To force to use and test Shapely 2.0, you have to set the environment variable USE_PYGEOS=0. You can do this before starting the Python process, or in your code before importing geopandas:

import os
os.environ['USE_PYGEOS'] = '0'
import geopandas

In a future release, GeoPandas will switch to using Shapely by default. If you are using PyGEOS directly (calling PyGEOS functions on geometries from GeoPandas), this will then stop working and you are encouraged to migrate from PyGEOS to Shapely 2.0 (https://shapely.readthedocs.io/en/latest/migration_pygeos.html).
  import geopandas as gp
```
So, being the type to look to the future, I set `USE_PYGEOS` to '0'. But I'm following the dask-geopandas [tutorial on spatial partitioning](https://dask-geopandas.readthedocs.io/en/stable/guide/spatial-partitioning.html), and calling the `calculate_spatial_partitions` method on my dask geodataframe produces:
```
ValueError: Metadata inference failed in `lambda`.

You have supplied a custom function and Dask is unable to 
determine the type of output that that function returns. 

To resolve this please provide a meta= keyword.
The docstring of the Dask function you ran should have more information.

Original error is below:
------------------------
TypeError('One of the arguments is of incorrect type. Please provide only Geometry objects.')

Traceback:
---------
  File "/home/guy/anaconda3/envs/geocube/lib/python3.10/site-packages/dask/dataframe/utils.py", line 182, in raise_on_meta_error
    yield
  File "/home/guy/anaconda3/envs/geocube/lib/python3.10/site-packages/dask/dataframe/core.py", line 6375, in _emulate
    return func(*_extract_meta(args, True), **_extract_meta(kwargs, True))
  File "/home/guy/anaconda3/envs/geocube/lib/python3.10/site-packages/dask_geopandas/core.py", line 140, in <lambda>
    pygeos.geometrycollections(part.geometry.values.data)
  File "/home/guy/anaconda3/envs/geocube/lib/python3.10/site-packages/pygeos/decorators.py", line 80, in wrapped
    return func(*args, **kwargs)
[...]

PR #232, opened 2023-01-22
COMPAT: allow using dask-geopandas with shapely 2.0 without pygeos

jorisvandenbossche:
Addresses #226

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,6 @@ install_requires = [
     "geopandas>=0.10",
     "dask>=2021.06.0",
     "distributed>=2021.06.0",
-    "pygeos",
     "packaging",
 ]
 
@@ -37,6 +36,7 @@ setup(
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python :: 3.11",
     ],
     keywords="dask geopandas spatial distributed cluster",
     description="Parallel GeoPandas with Dask",
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -64,4 +64,10 @@ jobs:
         run: |
           pytest -v -r a --color=yes --cov=dask_geopandas --cov-append --cov-report term-missing --cov-report xml .
 
+      - name: Test without PyGEOS
+        env:
+          USE_PYGEOS: 0
+        run: |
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -14,7 +14,6 @@
     "geopandas>=0.10",
     "dask>=2021.06.0",
     "distributed>=2021.06.0",
-    "pygeos",
```
martinfleis, line 17: With this, you can potentially install dask-gepandas with shapely 1.8 and no pygeos. I know that we cannot do shapely2 or pygeos in requirements but we shouldn't allow pip to resolve environment that is not sufficient to run it. 

That said, I don't know how to do that given the flexibility we have now so it may in the end be the best solution to keep it as you suggest and let people figure out potential issues from docs. 

Btw, how do you want to set requirements in conda-feedstock? The same and assume it will resolve with shapely 2?

martinfleis (approved):
Meant to hit approve since the only comment I made is in practice not resolvable now.

Thanks

setup.py
jorisvandenbossche: We could still do a run-time check, and error (or warn) if you have only shapely<2 without pygeos?

> Btw, how do you want to set requirements in conda-feedstock? The same and assume it will resolve with shapely 2?

I assume so, we also removed "pygeos" from the pyogrio feedstock (although for pyogrio, it is less critical to have pygeos or shapely>=2)

martinfleis: > We could still do a run-time check, and error (or warn) if you have only shapely<2 without pygeos

That may be wise. Otherwise we may run into some weird behaviour and possibly unclear error messages.

jorisvandenbossche: Ah, yes, that's true, I forgot that right now you actually get an error in `calculate_spatial_partitions` in such a case. While is not essential functionality (you can use dask-geopandas without ever calling this), we can provide a better error messaging (and since pygeos always has been required in practice by importing it top-level, it's not actually a big change to check for this in `__init__.py`)

jorisvandenbossche: OK, added a more informative error message. 

I don't think it is worth adding an extra CI build to cover this, but tested locally in an env without pygeos and with older shapely:

```
In [1]: import dask_geopandas
...
ImportError: Running dask-geopandas requires either Shapely >= 2.0 or PyGEOS to be installed. However, PyGEOS is not installed and Shapely is only at version 1.8.0
```

martinfleis:
Thanks! Agree there's no need for a CI env to check this. LGTM

Issue #237, opened 2023-02-09
Pickling DaskGeoDataFrame loses `spatial_partitions`


TomAugspurger:
This is a bit of a strange use-case, but roundtripping a GeoDataFrame through pickle loses the `spatial_partitions` property:

```python
import dask_geopandas
import geopandas.datasets
import pickle


df = geopandas.read_file(geopandas.datasets.get_path('naturalearth_lowres'))
ddf = dask_geopandas.from_geopandas(df, npartitions=2)
print(ddf.spatial_partitions)  # None
ddf2 = pickle.loads(pickle.dumps(df))
print(ddf2.spatial_partitions)  # AttributeError
```

`__init__` doesn't run, so it's never set at https://github.com/geopandas/dask-geopandas/blob/bc3af63aad32f939a0ac44950d39cc0f2da3b614/dask_geopandas/core.py#L67.

Maybe adding spatial partitions to `_args` to mirror dask.dataframe._Frame like in https://github.com/dask/dask/blob/main/dask/dataframe/core.py#L233-L241 would do the trick.

PR #238, opened 2023-02-09
Ensure spatial_partitions are properly pickled

TomAugspurger:
Closes https://github.com/geopandas/dask-geopandas/issues/237

### DIFF
--- a/dask_geopandas/core.py
+++ b/dask_geopandas/core.py
@@ -105,6 +105,17 @@ class _Frame(dd.core._Frame, OperatorMethodMixin):
                 )
         self._spatial_partitions = value
 
+    @property
+    def _args(self):
+        # Ensure we roundtrip through pickle correctly
+        # https://github.com/geopandas/dask-geopandas/issues/237
+        return super()._args + (self.spatial_partitions,)
+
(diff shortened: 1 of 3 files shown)


dask_geopandas/core.py
```diff
@@ -105,6 +105,17 @@ def spatial_partitions(self, value):
...
 
+    @property
+    def _args(self):
+        # Ensure we roundtrip through pickle correctly
+        # https://github.com/geopandas/dask-geopandas/issues/237
+        return super()._args + (self.spatial_partitions,)
+
+    def __setstate__(self, state):
+        *dask_state, spatial_partitions = state
+        super().__setstate__(dask_state)
```
TomAugspurger, line 116: Initially, I just copied over the `__setstate__` from dask.dataframe and added our state. By doing it like this (calling into `super()`) dask_geopandas should continue to work if Dask adds additional attributes to its state.

jsignell (approved):
This looks super reasonable to me and I really appreciate the attention to future compatibility.

martinfleis (approved):
Thanks! @TomAugspurger does this block anything on your side? We can potentially cut 0.3.1.

PR #244, opened 2023-03-20
typo fix

v2thegreat:
Order for the `points_from_xy` example was wrong. `points_from_xy` expects lon-lat, not lat-lon.

I ran was stuck on this for a bit until I realized what's going on😅


martinfleis (approved):
whoops. thanks!

Issue #245, opened 2023-03-24
FutureWarning for index_parts parameter in GeoDataFrame.explode()


ebkurtz:
I am calling the `.explode()` method on a dask GeoDataFrame which results in the warning:

>FutureWarning: Currently, index_parts defaults to True, but in the future, it will default to False to be consistent with Pandas. Use `index_parts=True` to keep the current behavior and True/False to silence the warning.

I tried passing `index_parts=False` which results in an error:
> TypeError: _Frame.explode() got an unexpected keyword argument 'index_parts'

Is it possible to have the method accept an `index_parts` argument to silence the warning? I believe this is the relevant piece of code, line 377 in core.py:

```
@derived_from(geopandas.geodataframe.GeoDataFrame)
def explode(self):
    return self.map_partitions(self._partition_type.explode, enforce_metadata=False)
```

Willing to help out with code changes, just wanted to see what the approach should be.

martinfleis:
Yes, we should mirror the api from geopandas. Which means exposing additional keywords here.

ebkurtz:
Great, I will make a fork and attempt a fix.

jorisvandenbossche:
Closed by https://github.com/geopandas/dask-geopandas/pull/246

PR #246, opened 2023-03-30
ENH: expose additional kwargs in explode method to mirror GeoPandas API

ebkurtz:
Expose `index_parts`, `column`, and `ignore_index` parameters in the `.explode()` method to mirror GeoPandas API. This allows supressing FutureWarning for `index_parts`, see  #245


TomAugspurger:
Can you define it on `Series` and `DataFrame` separately like in Dask?

- Series: https://github.com/dask/dask/blob/09dbeebbc1c8d5e40f7afc08258572ee897f36e9/dask/dataframe/core.py#L4050-L4053
- DataFrame: https://github.com/dask/dask/blob/09dbeebbc1c8d5e40f7afc08258572ee897f36e9/dask/dataframe/core.py#L5461-L5464

TomAugspurger:
The CI failure looks unrelated. I'll take a look at that later.

ebkurtz:
Thanks for the tip on defining them separately. I'll look at creating a unit test.

dask_geopandas/core.py
```diff
@@ -553,6 +549,15 @@ class GeoSeries(_Frame, dd.core.Series):
 
     _partition_type = geopandas.GeoSeries
 
+    @derived_from(geopandas.geodataframe.GeoSeries)
```
jorisvandenbossche, line 552:
```suggestion
    @derived_from(geopandas.GeoSeries)
```

dask_geopandas/core.py
```diff
@@ -807,6 +812,16 @@ def spatial_shuffle(
 
         return sorted_ddf
 
+    @derived_from(geopandas.geodataframe.GeoDataFrame)
```
jorisvandenbossche, line 815:
```suggestion
    @derived_from(geopandas.GeoDataFrame)
```

jorisvandenbossche (commented):
Thanks for the PR! The code looks good, would indeed be good to add a basic test using one of the keywords.

ebkurtz:
Would I add that as a new function in `test_core.py` are as additional code in the existing functions? I haven't done much with unit tests.

TomAugspurger:
Yes, a new function. You might be able to adapt a geopandas test like https://github.com/geopandas/geopandas/blob/dc695ee93113af5d1a93a5740ecd792c2b9d19b3/geopandas/tests/test_geom_methods.py#L1003.

ebkurtz:
Getting a lot of test failures after refreshing my fork. Looks like something with pyarrow strings?

jorisvandenbossche:
> Getting a lot of test failures after refreshing my fork. Looks like something with pyarrow strings?

Apologies for the slow follow-up. Indeed those failures were cased by something else, which has been fixed in the meantime. 

I also added the equivalent change to the new dask-expr based implementation.

jorisvandenbossche (approved):

jorisvandenbossche:
> The unit tests for the `ignore_index` argument are failing.

Your fix to create the expected result manually with the different index was still failing for the older geopandas or dask version. But so just added a skip for those older versions, as I don't think it is worth complicating the test for that.

jorisvandenbossche:
@ebkurtz Thanks again for the PR and apologies it took so long to get merged!

PR #242, opened 2023-03-17
`use_nullable_dtypes` arg removed in `dask`

j-bennet:
In https://github.com/dask/dask/pull/10076, `use_nullable_dtypes` will be removed from `_create_dd_meta` signature. Making a corresponding change here.

PR #251, opened 2023-04-23
[compat]: Fixed test for pandas>2

TomAugspurger:
pandas 2.0 changed a behavior w.r.t. groupby and non-numeric columns. This test doesn't need to drop the columns anymore. xref https://github.com/geopandas/geopandas/issues/1999

### DIFF
--- a/continuous_integration/envs/310-latest.yaml
+++ b/continuous_integration/envs/310-latest.yaml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   # required dependencies
   - python=3.10
-  - dask=2022.10.0
+  - dask
   - distributed
   - geopandas
   - pygeos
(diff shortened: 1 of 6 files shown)


TomAugspurger:
The failures in https://github.com/geopandas/dask-geopandas/actions/runs/4778643792/jobs/8495193757?pr=251 are fixed by #242 (confirmed by cherry picking and running locally)

Several of the other tests failed because of a pin on dask, but not pandas. Older versions of dask are not compatible with pandas>=2.0. I've pinned pandas in the `38-minimal` environment and unpinned dask elsewhere. So we should just see the parquet failures.

TomAugspurger:
@jorisvandenbossche I'd recommend merging #242 first, which should be good to go, and then I'll update this to confirm that CI is passing.

martinfleis (approved):
Thanks!

I suppose we should cut a patch release after this...

PR #265, opened 2024-02-12
TST: skip tests that need pygeohash/hilbertcurve/pymorton if not installed

jorisvandenbossche:
Given those are optional test dependencies and only used in a few tests, I think it's nice to skip those when you don't have them installed locally.


martinfleis:
Shall we remove those packages from `38-no-optional-deps.yaml`?

jorisvandenbossche:
Yes, good idea

martinfleis (approved):
Thanks!

PR #266, opened 2024-02-12
TST: use fixtures for accessing geopandas datasets

jorisvandenbossche:
This is a (potentially temporary) solution for the geopandas datasets being deprecated / removed. 

All usages have been updated to use a fixture, so we have a centralized place where the path is defined. And for now, just let that path be the old `geopandas.datasets` one, or the new path to the test data included in geopandas for newer versions. 
Longer term, we probably ideally wouldn't rely on such internal test data of geopandas, but that's something we can clean up later.


martinfleis:
This will not work. We no longer ship the data with the package, so it is not available in installed geopandas.

jorisvandenbossche:
Yeah, I figured that in the meantime .. It's only the dev build here, though, and makes that I can at least already run the tests locally (with a geopandas dev build)

What do we prefer? Include the data here as well on the short term? Longer term try to replace the cities/lowres with a simple generated points and polygons dataset?

martinfleis (approved):
Thanks!

PR #271, opened 2024-02-26
MAINT: move static metadata to pyproject.toml

jorisvandenbossche:
Some more modernization of the package infrastructure, mimicking what we have in geopandas.

For setuptools, I haven't yet added anything to pyproject.toml, still have to check what the default package discovery already gives.


martinfleis (approved):
Thanks!

PR #272, opened 2024-02-26
Update versioneer

jorisvandenbossche:
Re-run `versioneer install --vendor` with the latest version of versioneer. 

In addition also updated pre-commit to ignore top-level files (but only the package code itself). If we move config later for black/ruff we might make this smarter, but for now I want to avoid running black on the vendored `versioneer.py` file

And updating the release workflow to also run on the main branch


jorisvandenbossche:
This is a simple update, so going to merge this as it is blocking other PRs like https://github.com/geopandas/dask-geopandas/pull/271

PR #279, opened 2024-02-28
DOC: fix the RTD configuration file

martinfleis:
RTD is not building. This should fix it.


martinfleis:
All good https://readthedocs.org/projects/dask-geopandas/builds/23586759/

PR #280, opened 2024-02-28
Remove PyGEOS support, require Shapely 2+

jorisvandenbossche:
We don't actually have that much of code complexity from supporting both (at least not compared to pandas, because we mostly dispatch to the geopandas methods), but it does complicate testing a bit (e.g. when tests fail for pygeos), and generally I think shapely 2.0 is widely supported nowadays and already out for more than a year.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -46,7 +46,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Setup Conda
-        uses: conda-incubator/setup-miniconda@v2
+        uses: conda-incubator/setup-miniconda@v3
         with:
           environment-file: ${{ matrix.env }}
           miniforge-version: latest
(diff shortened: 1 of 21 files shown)


TomAugspurger:
Thoughts on adding `shapely>=2` to the dependencies in pyproject.toml? I suppose we get shapely through geopandas, but probably best to state it explicitly so that users don't need an ImportError to discover that their version isn't valid.

martinfleis:
> If there are multiple geometry columns, are they guaranteed to have the same CRS? It seems like there's a CRS on the table but not a GeoSeries.

Yeah, we need that as the min geopandas is 0.12 which does not enforce shapely 2.

jorisvandenbossche:
Good point to add shapely, done!

TomAugspurger:
+1. The doc build failure looks related to dask-expr, which is being handled elsewhere.

Issue #270, opened 2024-02-21
Error when reading geoparquet file


osjahputera:
I am using a small geoparquet file below (33 MB ~ building footprints in Singapore (SGP)), but the intention is to process larger geoparquet files covering much larger countries.

https://data.source.coop/vida/google-microsoft-open-buildings/geoparquet/by_country/country_iso=SGP/SGP.parquet

I am getting this error when reading the `SGP.parquet` file using the following code:

```
import dask.dataframe
import dask
import dask_geopandas
import geopandas

ddf = dask_geopandas.read_parquet(path='SGP.parquet')
```

<details> 

<summary> Error messages: </summary>

```
 ---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File ~\anaconda3\envs\geo_env\Lib\site-packages\dask\backends.py:141, in CreationDispatch.register_inplace.<locals>.decorator.<locals>.wrapper(*args, **kwargs)
    140 try:
--> 141     return func(*args, **kwargs)
    142 except Exception as e:

File ~\anaconda3\envs\geo_env\Lib\site-packages\dask\dataframe\io\parquet\core.py:529, in read_parquet(path, columns, filters, categories, index, storage_options, engine, use_nullable_dtypes, dtype_backend, calculate_divisions, ignore_metadata_file, metadata_task_size, split_row_groups, blocksize, aggregate_files, parquet_file_extension, filesystem, **kwargs)
    527     blocksize = None
--> 529 read_metadata_result = engine.read_metadata(
    530     fs,
    531     paths,
    532     categories=categories,
    533     index=index,
    534     use_nullable_dtypes=use_nullable_dtypes,
    535     dtype_backend=dtype_backend,
    536     gather_statistics=calculate_divisions,
    537     filters=filters,
    538     split_row_groups=split_row_groups,
    539     blocksize=blocksize,
    540     aggregate_files=aggregate_files,
    541     ignore_metadata_file=ignore_metadata_file,
    542     metadata_task_size=metadata_task_size,
    543     parquet_file_extension=parquet_file_extension,
    544     dataset=dataset_options,
    545     read=read_options,
    546     **other_options,
    547 )
    549 # In the future, we may want to give the engine the
    550 # option to return a dedicated element for `common_kwargs`.
    551 # However, to avoid breaking the API, we just embed this
    552 # data in the first element of `parts` for now.
    553 # The logic below is inteded to handle backward and forward
[...]

PR #290, opened 2024-05-06
BUG: fix read_parquet for file without CRS key in metadata (using implicit default OGC:CRS84)

jorisvandenbossche:
Closes https://github.com/geopandas/dask-geopandas/issues/270


martinfleis (commented):
Thanks!

martinfleis (approved):

PR #293, opened 2024-05-28
ENH: Add `from_wkt` and `from_wkb` tools 

### LINKED ISSUE #208: Create Dask GeoDataFrame from wkt string
I'm trying to load geometries from wkt.strings stored in a `dask.Dataframe` . I tried both in native `dask` and `dask_geopandas`, but the libraries seem not to be able to deal with wkt strings. Since the `dask_geopandas` community is probably more familiar with this type of issue, I thought I could better raise it here. 

Please find below a toy script with some things I've tried. 

```[python]
import dask.dataframe as dd
import dask_geopandas
import numpy as np
import shapely


def point():
    return shapely.geometry.Point(np.random.randint(low=0, high=90, size=2))


df = pd.DataFrame([wkt.dumps(point()) for i in range(100)], columns=["geometry"])
ddf = dd.from_pandas(df, npartitions=4)
# ddf.compute().geometry.apply(wkt.loads)  # runs fine
# ddf.geometry.apply(wkt.loads)  # raises ParseException: Unknown type: 'FOO'


dask_geopandas.from_dask_dataframe(ddf, geometry="geometry")  # TypeError("Input must be valid geometry objects: {0}".format(geom))


def partition_wrapper(df):
    df.geometry = df.geometry.apply(wkt.loads)
    return df

dask_geopandas.from_dask_dataframe(ddf).map_partitions(partition_wrapper)  # TypeError("Only str is accepted.")
```

What would be the suggested work around? Or should a functionality to load from wkt strings become available in `dask_geopandas`?

giorgiobasile:
Implements `dask_geopandas.from_wkt` and `dask_geopandas.from_wkb`, taking inspiration from `dask_geopandas.points_from_xy`.

Closes #208


jorisvandenbossche:
@giorgiobasile Thanks for the PR!

One question we have to think about is the exact API: do we want to do this at the dataframe level, or at the (geo)series level? Right now you added a function that takes a dataframe and returns a GeoSeries. But if it returns a GeoSeries, we could also let it accept a Series (and the user can do the column selection themselves). Or otherwise let it accept and return a (Geo)DataFrame.

giorgiobasile:
Thanks @jorisvandenbossche for your remarks! 

> One question we have to think about is the exact API: do we want to do this at the dataframe level, or at the (geo)series level?

Yes, the reason I sticked with the `dask.dataframe` -> `gpd.GeoSeries` interface was mainly for consistency with `dask_geopandas.points_from_xy` which I took inspiration from, as I needed to do something like:

```python
gdf = gpd.from_dask_dataframe(
    ddf, geometry=dask_geopandas.from_wkt(ddf, geometry_wkt='wkt', crs="EPSG:4326")
)
```

Just double checking I'm getting what you are suggesting:

> But if it returns a GeoSeries, we could also let it accept a Series (and the user can do the column selection themselves)

So this would change to allow for something like:

```python
gdf = gdp.from_dask_dataframe(
    ddf, geometry=dask_geopandas.from_wkt(ddf["wkt"], crs="EPSG:4326")
)
```

> Or otherwise let it accept and return a (Geo)DataFrame.

In this case the user would have a unified API, like:

```python
gdf = dask_geopandas.from_wkt(ddf, geometry_wkt='wkt', crs="EPSG:4326")
```

Does this all make sense? Also, If we go for any of the latter two cases, should the interface be applied to `dask_geopandas.points_from_xy` as well, for sake of consistency?

jorisvandenbossche:
Apologies for the slow follow-up!

> Yes, the reason I sticked with the `dask.dataframe` -> `gpd.GeoSeries` interface was mainly for consistency with `dask_geopandas.points_from_xy` which I took inspiration from

Yes, that's a good point. `points_from_xy` has the constraint that it has multiple input columns, though, which we don't have for `from_wkb`, so we have a bit more flexibility.




> > But if it returns a GeoSeries, we could also let it accept a Series (and the user can do the column selection themselves)
> 
> So this would change to allow for something like:
> 
> ```python
> gdf = dask_geopandas.from_dask_dataframe(
>     ddf, geometry=dask_geopandas.from_wkt(ddf["wkt"], crs="EPSG:4326")
> )
> ```

I think this feels the most natural API to me / most consistent with how it works in geopandas.

giorgiobasile:
> Apologies for the slow follow-up!

No problem!

> I think this feels the most natural API to me / most consistent with how it works in geopandas.

Ok sure, I just updated it accordingly.

jorisvandenbossche (approved):
Thanks for the update, looks good!

jorisvandenbossche:
Thanks @giorgiobasile!

giorgiobasile:
> Thanks @giorgiobasile!

@jorisvandenbossche thank you for your suggestions!

PR #294, opened 2024-06-01
MAINT: update linting setup (align with geopandas, use ruff, move config to pyproject.toml)

jorisvandenbossche:
Mostly just copied geopandas' setup (with its latest changes to expand the usage of ruff in https://github.com/geopandas/geopandas/pull/3291 and https://github.com/geopandas/geopandas/pull/3297), with some minor edits to adjust for dask-geopandas

The diff is mostly from sorting imports, and only few actual code changes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,18 @@
+ci:
+    autofix_prs: false
+    autoupdate_schedule: weekly
+
 files: 'dask_geopandas\/'
 repos:
-  - repo: https://github.com/psf/black
-    rev: 22.12.0
-    hooks:
-      - id: black
-        language_version: python3
-  - repo: https://github.com/pycqa/flake8
-    rev: 5.0.4
-    hooks:
-      - id: flake8
-        language_version: python3
+    - repo: https://github.com/psf/black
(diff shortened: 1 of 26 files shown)


.pre-commit-config.yaml
```diff
@@ -1,12 +1,18 @@
+ci:
+    autofix_prs: false
+    autoupdate_schedule: weekly
```
martinfleis, line 3: Can we do quarterly? Ruff's release frequency otherwise generates update PRs a bit too often to my liking.

jorisvandenbossche: Certainly (I just copied it from geopandas)

jorisvandenbossche:
```suggestion
    autoupdate_schedule: quarterly
```

martinfleis: Ah. We should update it there as well. In any case, we shall enable geopandas org on pre-commit-ci, right now this does nothing :)

jorisvandenbossche: > right now this does nothing :)

Ah, that's why I couldn't relate with your "generates update PRs a bit too often to my liking" ;)

PR #295, opened 2024-06-03
Register arrow import/export dispatch to make p2p shuffle work

jorisvandenbossche:
This registers our own implementation for pyarrow table <-> (Geo)DataFrame conversion through the dispatch mechanism. The distributed p2p shuffle algorithm uses that under the hood (it converts to pyarrow to send the data to other workers), and so with this change (and with the latest distributed release, as we require the fix from https://github.com/dask/distributed/pull/8667), a distributed `spatial_shuffle()` now works.

### DIFF
--- a/dask_geopandas/backends.py
+++ b/dask_geopandas/backends.py
@@ -1,6 +1,7 @@
 import uuid
 from packaging.version import Version
 
+import dask
 from dask import config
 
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Going to merge this, as it is required for me to get https://github.com/geopandas/dask-geopandas/pull/165 running with the latest versions of dask/distributed

PR #298, opened 2024-06-24
RLS/BLD: fix release workflow to produce correctly versioned artifacts

jorisvandenbossche:


jorisvandenbossche:
I went the dirty way of including this in the actual 0.4.0 release commit and repushed the tag, to avoid having to make a new 0.4.1 release for this change that doesn't impact that actual package source.

Issue #303, opened 2024-07-18
`ValueError: 'left_df' should be GeoDataFrame, got <class 'tuple'>` using sjoin after `spatial_shuffle`


timdhu:
Hello,

I'm getting an error when I try to use `sjoin` after using `spatial_shuffle`. 

Here are the packages in my environment:

```
Package            Version
------------------ -----------
certifi            2024.7.4
charset-normalizer 3.3.2
click              8.1.7
cloudpickle        3.0.0
dask               2024.7.0
dask-expr          1.1.7
dask-geopandas     0.4.1
distributed        2024.7.0
fsspec             2024.6.1
geodatasets        2024.7.0
geopandas          1.0.1
idna               3.7
importlib_metadata 8.0.0
Jinja2             3.1.4
locket             1.0.0
MarkupSafe         2.1.5
msgpack            1.0.8
numpy              2.0.0
packaging          24.1
pandas             2.2.2
partd              1.4.2
pip                24.0
platformdirs       4.2.2
pooch              1.8.2
psutil             6.0.0
pyarrow            17.0.0
pyogrio            0.9.0
pyproj             3.6.1
python-dateutil    2.9.0.post0
pytz               2024.1
PyYAML             6.0.1
requests           2.32.3
setuptools         69.2.0
shapely            2.0.5
six                1.16.0
sortedcontainers   2.4.0
tblib              3.0.0
toolz              0.12.1
tornado            6.4.1
tzdata             2024.1
urllib3            2.2.2
zict               3.0.0
zipp               3.19.2
```

(I ran `pip install dask-geopandas` and `pip install geodatasets` in a fresh venv)

Then I run the following code:

```
import geopandas
import geodatasets
import dask_geopandas

colombia = geopandas.read_file(geodatasets.get_path('geoda.malaria'))
colombia = dask_geopandas.from_geopandas(colombia, npartitions=4)
colombia = colombia.spatial_shuffle()
colombia.sjoin(colombia).compute()
```

and get the following error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../env/lib/python3.11/site-packages/dask_expr/_collection.py", line 476, in compute
    return DaskMethodsMixin.compute(out, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../env/lib/python3.11/site-packages/dask/base.py", line 376, in compute
[...]

PR #307, opened 2024-08-28
Handle dask-expr optimizations in sjoin

TomAugspurger:
Closes https://github.com/geopandas/dask-geopandas/issues/303

This is a simple fix. I think we might want to explore what a proper dask-expr Expression for this looks like. I'll look into that later.


martinfleis (approved):
Thanks!

PR #317, opened 2024-11-09
Include dask[diagnostics] in dev requirements

dluks:
When setting up the project for development, tests fail with the current `requirements-dev.txt` due to the dask diagnostics suite missing.

This includes the required diagnostics libraries (`bokeh` and `distributed`) to ensure all tests pass after a default dev environment setup.


martinfleis (approved):
thanks!

Issue #321, opened 2024-12-11
Dropping the geometry column no longer produces a Dask DataFrame object.


fbunt:
Dropping the geometry column from a dask-geopandas GeoDataFrame no longer produces a dask DataFrame object. This causes errors in future operations where a GeoDataFrame is not expected or where the resulting GeoDataFrame thinks it should have a geometry column. The issue seems to stem from the dask-expr backend.

Example:

```python
>>> import dask_geopandas as dgpd
>>> import geopandas as gpd
>>>
>>> gdf = gpd.GeoSeries.from_xy(x=[1, 0], y=[0, 1], crs=4326).to_frame("geometry")
>>> gdf["data"] = [1, 2]
>>> gdf = dgpd.from_geopandas(gdf, npartitions=1)
>>> df = gdf.drop(columns="geometry")
>>> print(type(df))
dask_geopandas.expr.GeoDataFrame
```

Example Error:
<details>

```python
>>> df.shape
<class 'dask_geopandas.expr.GeoDataFrame'>
Traceback (most recent call last):
  File "/home/fred/homes/rts/anaconda3/envs/rts-dev/lib/python3.12/site-packages/geopandas/geodataframe.py", line 517, in crs
    return self.geometry.crs
           ^^^^^^^^^^^^^
  File "/home/fred/homes/rts/anaconda3/envs/rts-dev/lib/python3.12/site-packages/pandas/core/generic.py", line 6299, in __getattr__
    return object.__getattribute__(self, name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fred/homes/rts/anaconda3/envs/rts-dev/lib/python3.12/site-packages/geopandas/geodataframe.py", line 253, in _get_geometry
    raise AttributeError(msg)
AttributeError: You are calling a geospatial method on the GeoDataFrame, but the active geometry column ('geometry') is not present. 
There are no existing columns with geometry data type. You can add a geometry column as the active geometry column with df.set_geometry. 

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/fred/homes/rts/anaconda3/envs/rts-dev/lib/python3.12/site-packages/dask_expr/_core.py", line 493, in __getattr__
    return object.__getattribute__(self, key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #322, opened 2024-12-21
Fixed dropping the geometry column

TomAugspurger:
Closes https://github.com/geopandas/dask-geopandas/issues/321.

We need to override the `Drop` from dask-expr, since we potentially need to change the return type from a GeoDataFrame to a regular DataFrame.


TomAugspurger:
Not sure about the codecov failure, but all the new code is tested.

PR #326, opened 2025-03-13
remove geom_almost_equals

martinfleis:
fixes dev CI and downstream dev CI by following geopandas in removal of `geom_almost_equals`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,16 @@
 Changelog
 =========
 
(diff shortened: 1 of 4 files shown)


martinfleis:
Hmm, this has actually uncovered more failures as the tests collection no longer errors. But I'd rather merge this and leave the other failures for later as this now causes downstream CI that is testing against geopandas and dask-geopandas mains to fail.

TomAugspurger (approved):

PR #334, opened 2025-08-11
GHA: use micromamba

martinfleis:
Copy from geopandas.


martinfleis:
Given that main works and libmamba seems to struggle to resolve the environments here, closing.

PR #347, opened 2026-04-08
BUG: align CI Dask dependencies with dataframe API requirements

rithiksachdeva:
## Summary

Fix CI failures caused by CI environments installing Dask variants that do not provide the dataframe API required by the current codebase.

## Root Cause

Some CI environments install `dask` or `dask-core` without guaranteeing dataframe support, while the project now requires `dask[dataframe]>=2025.1.0`.

This causes test collection to fail with:

`ModuleNotFoundError: No module named 'dask.dataframe.api'`

## Fix

- Replace CI environment dependencies that use `dask` or `dask-core`
- Use `dask[dataframe]>=2025.1.0` instead
- Align CI environments with the dependency baseline declared in `pyproject.toml`

## Scope

This PR only updates CI environment dependencies. It does not change library behavior, join semantics, partition planning, or test logic.

## Result

- Restores test collection in CI
- Ensures CI matches the project’s declared Dask dependency requirements
- Unblocks feature PRs that were failing due to the same environment mismatch

### DIFF
--- a/ci/envs/310-minimal.yaml
+++ b/ci/envs/310-minimal.yaml
@@ -5,6 +5,7 @@ dependencies:
   # required dependencies
   - python=3.10
   - numpy=1.24
+  - dask=2025.1.0
   - distributed=2025.1.0
   - geopandas=0.14.3
   - pandas=2.0.0
@@ -20,5 +21,4 @@ dependencies:
   - pyarrow
   - pip
   - pip:
-      - dask[dataframe]>=2025.1.0
       - pymorton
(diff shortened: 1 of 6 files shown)


martinfleis (commented):
Thanks for the fix but please make sure dask and its dependencies needed for data frame support are pulled from conda-forge, not from PyPI.

rithiksachdeva:
Closing this PR as this approach is no longer being pursued.

I explored this direction, but I’m taking a different approach going forward. Thanks for the feedback!

PR #352, opened 2026-06-23
DEPS: update required dependencies and extend CI to test python 3.13. 3.14

martinfleis:
An attempt to fix CI, that has been failiing for some time because some environments were pulling old Dask. 

Adding Python 3.13 and 3.14 to the mix at the same time. We should also bump the minimal one but let's focus on green CI now.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -25,6 +25,7 @@ jobs:
       run:
         shell: bash -l {0}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest]
         env:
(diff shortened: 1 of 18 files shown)


martinfleis:
I have bumped deps to match current GeoPandas in the end. It proved difficult to find env spec that would resolve otherwise.

jorisvandenbossche:
Thanks for looking into this!

So essentially this PR bumps the minimum required versions for GeoPandas from 0.14.3 to 1.0 and shapely from 2.0 to 2.1? And also minimum pandas to 2.0? (which I think itself is not enforced by geopandas 1.0)

And also cleans up a bunch of old compat stuff (eg for older dask versions) that was no longer needed (even before this PR)?

jorisvandenbossche (approved):

martinfleis:
Yes to all.

